PR #16, opened 2019-05-05
VP-110: Set fixed Mongo version for mockgoose to avoid test hangs on …

thjc:
…Linux


thjc:
This is needed for full set of tests to pass on my system, not sure if there are alternate fixes.

avowkind:
confirmed tests still pass on Macos

PR #20, opened 2019-05-07
Switch to slimmer production docker image with only built files

thjc:
There is still a lot of stuff that is probably not needed in node_modules, but this helps to cut the final production docker size a little bit.


avowkind:
What command are you using to run the prod instance?   Docker or docker-compose?
A>

PR #32, opened 2019-05-14
Register interest server side

darcycox97:
- This PR should handle all the required server-side functionality for adding "interests" to the db. 
- An interest consists of two required "ObjectId"s which reference an opportunity and a person. It also contains an optional comments property.
- Uses a 3rd party mongoose plugin to ensure the interest is only added if the referenced ids exist in the db.
- Tests reflect both the positive and negative cases for adding valid / invalid interests to the db.
- On my machine, the tests pass, but there are uncaught exceptions because the DB connection is reset as I mentioned


avowkind:
Thanks for the excellent PR. 100% test coverage and a nice clear set of functions. All tests passed first time.   Recommend adding 'standard' to your work environment and enforcing layout standards such as tabs, quotes and semicolons.

PR #40, opened 2019-05-21
IGNORE PR - Vp 127 test docker builds failing trying to download mong

avowkind:
Testing PR and travis


avowkind:
Dropped out Travis for now as CirrusCI is running the tests and also caches npm modules so builds run a _lot_ faster.

PR #41, opened 2019-05-21
Vp 131 use redux instead of local state in oplistsection

avowkind:
removed what was a bad example of usage. If using redux api wrappers then the results are put in the store and you can access them from there. no need to use local state.


avowkind:
Closing - too many changes here

PR #44, opened 2019-05-22
VP-102: use package-lock files for docker builds

thjc:


thjc:
I think there is a better command for this

PR #45, opened 2019-05-22
VP-102: Make sure we include package-lock for docker build and use np…

thjc:
…m ci instead of install


thjc:
package lock is included as mongodb-memory-db needed to match the tighter version spec in the package.json file

avowkind:
I propose closing this PR.  I think the equivalent changes have made their way to master by other means. At least using npm ci in the cirrus file and not excluding package-lock.json from the docker files.

thjc:
Fixed elsewhere now

PR #48, opened 2019-05-23
Vp132   footer

waltzaround:
footer and about page touchup


waltzaround:
Also fixed some responsive issues on landing page

PR #64, opened 2019-05-25
VP-102: Make sure all the tests run in cirrus

thjc:


thjc:
Tests will fail since one of the not run tests on master is failing (as far as I know)

PR #80, opened 2019-05-26
Vp 32

AidanTheDeveloper:
Status made on VP-32, images are uploading to a file location, although they are not yet being displayed.


avowkind:
The Uploader is failing in OpDetailForm.spec.js. It looks like the uploader component creates a child called IframeUploader which in turn tries to call win.document. which doesn't exist as the test is not a browser. That may make it very hard to test. although you can get a mocker for the browser window https://github.com/lukechilds/window. Before doing that I'd try one of the other Uploader patterns that doesn't use Dragger to see if its a localised issue. If it works by creating an IFrame that may be hard to test.

PR #83, opened 2019-05-26
VP-102: Automatically deploy to alpha on successful master build

thjc:


thjc:
I imagine this is going to need more tweaking, but its a start...

PR #90, opened 2019-05-27
add tests for nterestTable, exclude helpers nyc

avowkind:


avowkind:
The uncaught exceptions need to be investigated but are not enough to stop the merge.  Could you please approve.

Thanks Andrew

PR #93, opened 2019-05-28
Vp 47 - added dummy volunteer activity dashboard

waltzaround:


waltzaround:
Still needs logic and more polish but the skeleton is there

PR #95, opened 2019-05-29
Vp 168 uncaught exceptions when running tests on windows

avowkind:
Cause:  mongoose disconnect. 

we are running tests in parallel.  Although we now use mongo-memory-server to give us multiple parallel databases mongoose is still a single thing.  So mongoose connect connects to different ports but when we call mongoose.disconnect it drops everything.   
Commenting out and the tests run ok.  there may be dangling connections but the tests don't care. 

Tested on Windows.

### DIFF
--- a/server/api/person/__tests__/person.spec.js
+++ b/server/api/person/__tests__/person.spec.js
@@ -4,20 +4,17 @@ import { server, appReady } from '../../../server'
 import Person from '../person'
 import MemoryMongo from '../../../util/test-memory-mongo'
 import people from '../__tests__/person.fixture'
-// Initial people added into test db
 
-let memMongo
-
-test.before('before connect to database', async () => {
+test.before('before connect to database', async (t) => {
   await appReady
-  memMongo = new MemoryMongo()
+  t.context.memMongo = new MemoryMongo()
   // console.log('App ready')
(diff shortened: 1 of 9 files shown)


server/api/person/__tests__/person.spec.js
```diff
@@ -4,20 +4,17 @@ import { server, appReady } from '../../../server'
...
-  await memMongo.start()
+  await t.context.memMongo.start()
 })
 
-test.after.always(async () => {
-  await memMongo.stop()
-  // console.log('stopped')
+test.after.always(t => {
+  t.context.memMongo.stop()
+  console.log('stopped')
```
michael-j-chinn: Should we comment out or remove this log?

michael-j-chinn:
What are the ramifications of those hanging connections? How long for them to get cleaned up? Is there any concern of resource exhaustion during heavy repeated test running?

server/api/person/__tests__/person.spec.js
avowkind: should remove eventually.

avowkind:
I think the connections will close fairly soon anyway after the test goes out of scope and the GC runs. As we don't have this problem on the mac I suspect it is just a timing issue.

PR #104, opened 2019-06-02
VP-41: Make New opportunity only available for op-providers

thjc:
I think this is correct behaviour, unless you want a non op-provider to be able to click the button and then have the opportunity to sigh in with a different account...


waltzaround:
will check it out in a second :)

waltzaround:
Just tried it out, awesomesauce @thjc 

yay

PR #126, opened 2019-06-12
Fix bug in OpDetail display

MinhPhu0304:
Markdonw will complain if the op.description is undefined becuase user left the description blank. Fixed by checking if the description is undefined we make is an empty string


MinhPhu0304:
Not ready yet since there is a warning of failed prop type

PR #138, opened 2019-06-21
Vp226 action cards

waltzaround:
added action cards to home

They lack logic on when to show though, so only drawing two generic ones:
- find something to do
- join development on github


waltzaround:
Little wins :)

PR #141, opened 2019-06-22
Add SMTP_ID and SMTP_PWD to docker env set at build time

thjc:
Should work, but a bit tricky to test without end to end...

Still needs actual values added to the cirrus build, I will add a wiki page detailing that. They should be added as env vars

BUILD_SMTP_ID and BUILD_SMTP_PWD in the .cirrus.yml file in the env block as ENCRYPTED values

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,10 @@ RUN npm run prod-build
 
 FROM node:12-alpine as production
 ARG BUILD_REVISION=local-build
+ARG BUILD_SMTP_ID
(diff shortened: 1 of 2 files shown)


thjc:
Wiki page created to describe adding the encrypted vars with the actual private values. Either add to this branch or create as separate PR

https://voluntarily.atlassian.net/wiki/spaces/VP/pages/26640387/Using+Encrypted+Env+Variables+with+Cirrus+CI

PR #145, opened 2019-06-23
Vp 70 change home redirect

waltzaround:
emergency fix for persona section


waltzaround:
css was broken on landing page in the persona component, this fixes it

PR #146, opened 2019-06-24
[VP-213] Add ability based on authenticated user to request

michael-j-chinn:
This new middleware will scan through APIs looking for authorization rules it can piece together into one big Ability. This Ability is then attached to the request object. By specifying Ability files per entity (i.e. opportunity.ability.js) we increase parallelization of work and keep rules close to their entity. Going forward we can use this Ability in various ways:

- Authorize API requests before they hit a controller using middleware to return a 401 Unauthorized response
- Add plugin to Mongoose that uses Ability to authorize DB call. Unauthorized calls get canceled before they hit the DB and return an empty array.
- Add plugin to Mongoose that allows Ability to specify Schema fields accessible at different authorization levels. Use this to filter any "select" params coming in on the request.
- Run middleware after the controller response that strips unauthorized fields from the response data

Aside:

I also want to start pushing us towards "enum" type objects instead of strings littered everywhere. I've done a bit in this PR to that end.


michael-j-chinn:
Note: Future PRs will do the same for the front end but be more granular. This PR is back-end focused.

PR #152, opened 2019-06-26
Vp 257 include tags in search

darcycox97:
Backend treats the search query as an array of keywords and will now include any opportunities with  one or more matching keywords, as well as any opps with a matching title, subtitle, or description (as before)


waltzaround:
Looks good, tested on my machine

PR #155, opened 2019-06-26
Tests/vp 217/registerinterestitem

avowkind:


darcycox97:
Just needs the removal of the extra html file and console.log statement, as we noticed in the review

PR #160, opened 2019-06-27
Complete Opportunity progress

AidanTheDeveloper:


AidanTheDeveloper:
Looks like there is a failing test, I will revisit this tomorrow

AidanTheDeveloper:
Hey Andrew, I tried running the server and the interests were not showing up. After looking into this I found that in the server/api/interest/interest.js file the field which contains the opportunity ID is "opportunity". After changing op to "opportunity" the interests started appearing. I notice in your test the expectation is for "op" to be requested from redux. Could you verify that interests are showing up for you in the opdetail page without the commit that I have made? Perhaps there is a change that has been made to the interest Schema that has yet to be merged into this branch.

AidanTheDeveloper:
Hey Andrew, for now I've modified the test to use "opportunity" in the mocked fetch. If we do want to use "op" instead feel free to revert the change.

AidanTheDeveloper:
I've looked into the op thing a bit more. Looks like the change I made causes the API to retrieve all interests instead of interests for one specific opportunity. 

I was having difficulty retrieving interests for a single opportunity and so thought the change I made had fixed the issue I was experiencing. It looks like there is probably another reason why interests are not being retrieved on my machine. I will revert the changes I have made and keep investigating the issue I am experiencing.

AidanTheDeveloper:
The issue I was experiencing was that I had not set up my mongo database correctly. I think I used the sample interest.json files and manually imported them into Robo3T. I did not realize that the opportunity field needed to be an ObjectID and not a String in order for it to be picked up interest controller.

PR #162, opened 2019-07-02
VP 233 Search by location

cyrus-raitava:
Introduces a searchable location input component to both the update page for opportunities, as well as the page for creating new opportunities. Inputs are constrained currently to New Zealand TLAs and regions, as is retrieved from mongoDB via a newly made endpoint. BE/FE tests have been included as well. Example of the component in action is below! (To test, check out updating an opportunity/creating a new opportunity). 

***

image


darcycox97:
Note - this does not fully implement VP 233, but we thought it was enough work to justify creating a PR

PR #172, opened 2019-07-04
Add location filter to search page

darcycox97:
- Added location to OpCard to make it more clear that the filter is working.
- Filtering is done server-side via the "location" query parameter. Will return all territories **within** a region, or just the matching territory if there are no sub-locations.
- Styling isn't perfect but should be a good start. @waltzaround  maybe you could take a look at improving this at a later date? CSS isn't my strong point


waltzaround:
can do @darcycox97 🔥

PR #174, opened 2019-07-06
[VP-287] Moving interests into interest-archive when opportunity is a…

AidanTheDeveloper:
…rchived


AidanTheDeveloper:
Whoops, forgot to actually add the interest-archive model

PR #185, opened 2019-07-13
FIx broken new activity link

darcycox97:
New activity was inaccessible due to incorrect routing


server/routes.js
```diff
@@ -25,7 +25,6 @@ module.exports = routes()
   // Activities
   .add('acts', '/acts', 'act/actlistpage')
   .add('act', '/acts/:id', 'act/actdetailpage')
-  .add('actedit', '/acts/:id/edit', 'act/actupdatepage')
-  .add('actnew', '/act/new', 'act/actupdatepage')
-  .add('actsection', '/act/section', 'act/actlistsection')
+  .add('actnew', '/act/:new', 'act/actdetailpage')
```
sezflynn: Hi, shouldn't the route for a new activity be `/acts/new` ? acts=plural because we're still adding a new activity to the activities collection.

Correct me if I'm wrong but doesn't `:new` indicate a placeholder for a variable?

PR #219, opened 2019-07-14
Vp 293 add tags to activities

darcycox97:
- adds tags to activities and considers them in the search endpoint


darcycox97:
@MinhPhu0304 re-created the PR because something went wrong with the other one

PR #220, opened 2019-07-14
VP388 - emergency fix for search results

waltzaround:


waltzaround:
We need faster tests :L

PR #223, opened 2019-07-16
Date filter unit test fix

MinhPhu0304:
For weird reason the value passed in filter date for OplistSection unit test is in incorrect ISO format. This PR will fix the issue


julia-mareike:
hey :D 
i had a look at this test and i think actually the way the dates are formatted for the test opportunities is what's causing it to fail. 

at the moment the dates are an array of objects with this format `{'$date': '2019-05-23T12:26:18.000Z'}` ie on line 25 in `OpListSection.spec.js` - but by the time this gets rendered by `OpCard` on line 21, `moment()` can't parse and format that object, and so it just defaults to using the current time - meaning it sets the startTime for all test opportunities as today. Which means when the `filterDateState` happens to be today's date, like it is today, it returns it as a found result!

if the date structure for hard-coded opportunities becomes `date: ['2019-05-23T12:26:18.000Z', '2019-06-12T04:55:10.014Z']` the right dates get rendered and the tests pass 👍

PR #229, opened 2019-07-17
Story/vp 367/activities vertical menu

waltzaround:


components/Act/ActDetail.js
```diff
@@ -66,7 +66,7 @@ export function ActDetail ({ act }) {
           <Spacer />
         </Left>
         <Right>
-          <img style={{ width: '100%' }} src={img} alt={act.title} />
+          <img style={{ width: '100%' }} src={act.imgUrl} alt={act.title} />
```
MinhPhu0304: I think this should be img variable instead. Because if the activity does not have image, we will use default image

MoniqueWarrington: I think Walts code is primarily styling? It might be @alisamhh who wrote the logic <3

MoniqueWarrington: I think there is a default fallback for if an activity/opportunity doesn't have an image

waltzaround: Yerp, @alisamhh wrote the logic

PR #275, opened 2019-07-27
fix for twitter and facebook placeholder display to not show when empty

vincefnz:
i have not updated tests to include facebook and twitter properties as they were not in current tests, is that ok?


waltzaround:
I'd recommend still adding the tests in

PR #292, opened 2019-08-05
VP-428 Add location attribute to people's profiles, and add recommended opportunities from this

cyrus-raitava:
This PR will allow for the addition of the location of a person (based on New Zealand Regions and Territories), as well as subsequent recommendations of opportunities based on that location! Proximity is taken into account (e.g opportunities in other territories in the same region will be in the list of recommended opportunities). 

## **Notable changes:**
- Addition of location attribute to Person Schema
- Restructuring of object returned by /location redux endpoint
- Introduction of OpRecommendations component into PersonHomePage

Any questions, hit up @darcycox97 or I!


darcycox97:
We are aware of the linting issue, but can't figure out how to resolve it. It appears there is a conflict between two rules so either way one of them will complain and the other won't.

avowkind:
I'll check out and look at the issues.

PR #301, opened 2019-08-09
Fix issue with recommended ops not showing

darcycox97:
This PR fixes the issue where recommended ops aren't showing when they should. 
The problem was due to the initial redux api call filtering the opportunities to only include those I have requested. So I removed this filter and implemented it in the front-end instead, so that the Recommendations component gets passed all ops


darcycox97:
@avowkind Have moved recommendations to server side as suggested, and updated tests for the new approach

PR #325, opened 2019-08-25
Vp 177 sign in should not stack in the history but be replaced so that back works as advertised

Eteroa123:
<h1> Pull Request  </h1>
<p> Vp 177 sign in should not stack in the history but be replaced so that back works as advertised</p>

<h2> Proposed Changes? </h2>
<p>1. When a user visits an activity and presses "The person" it is redirected to auth (Sign in /Sign up) </p>
<p>2. When user has signed up/in they will then be directed back to the page they were before</p>


Eteroa123:
@avowkind oh man I just saw it okay i'll fix that. 

Btw Can I create a new ticket and branch for this new fix

Eteroa123:
@avowkind  Actually no ill use this branch

PR #327, opened 2019-08-27
Add recommendations of opportunities based on a user's skills

cyrus-raitava:
## Proposed Changes
This PR will introduce new functionality to the component **(OpRecommendations)** into the /home page, to display recommended opportunities from the platform.

These will now either be based on:
* **Location** (e.g opportunities in the exact same region and/or related territories - e.g Kaikoura District and Christchurch City are both territories within the Canterbury region) - _(PREVIOUS FUNCTIONALITY)_
* **Skills** - depending on the number of matching skill tags a given opportunity has with a user, that opportunity gets assigned a value (+1 for each matching tags). These are then ranked based on calculated count, from most to least relevant - _(ADDED FUNCTIONALITY)_

Any questions about this, feel free to ask @darcycox97 or I!!

## Screenshots
image

### DIFF
--- a/components/Op/OpRecommendations.js
+++ b/components/Op/OpRecommendations.js
@@ -6,13 +6,20 @@ import OpList from './OpList'
 class OpRecommendations extends React.Component {
   render () {
     const { recommendedOps } = this.props
-    return (recommendedOps.basedOnLocation.length === 0)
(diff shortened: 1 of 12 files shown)


avowkind:
Also - branches MUST be named with the VP-number

PR #384, opened 2019-09-15
VP-451 - ajusting organisations to return fields name, imgUrl, about, category"

rogerioavsilva:
# Pull Request Template (VP-123)

Please check that your ticket number and ticket description is included in the PR title 😀

## Proposed Changes? 🤔
1. ajusting organisations to return fields name, imgUrl, about, category"

## Additional Info.🧐

Any additional info goes here


rogerioavsilva:
I couldn't make the test works, I need some help
If I put "p={"name": 1"}" on url, json is undefined and the test doesn't work.


// test.only('Should correctly give a list with only name field', async t => {
//   const nameOrganisation = "OMGTech";
//   p={"name": 1"} when applied on url, json is undefined
//   const res = await request(server)
//     .get('/api/organisations?q={"name":"' + nameOrganisation + '"}&p={"name": "1"}')
//     .set('Accept', 'application/json')
//     .expect(200)
//     .expect('Content-Type', /json/)
//   const got = res.body
//   console.log('got', got[0].name)
//   t.is(got[0].name, nameOrganisation)
//   t.is(got[0].about, null)
// })

PR #389, opened 2019-09-20
VP-297 create opportunity from an activity

avowkind:
## Proposed Changes? 🤔
1. Add fromActivity to the op record 
2. Create Opportunity on Act details page button wired to op/new?act=id
3. In Opdetailspage (new) init the op from the act if an activity id is listed.  Copy over the fields. 
4. extended test cover on opdetailspage

NewOpFromAct780p


waltzaround:
Reviewing now

PR #393, opened 2019-09-22
Feature: badge issuing system

MinhPhu0304:
VP-240: As a volunteer I want to track my status and achievements (Badges) 😀

## Proposed Changes? 🤔
1. Added new database schema for storing issued badge in from the system
2. Added front end component for displaying and issuing badge
3. API endpoint for listing badges available from Voluntarily issuers
4. API endpoint for issuing badge
5. Configuration check to hide badge system if badgr credential is not found
6. Unit test for badge API 
7. Using Hooks component instead of class component in the new badge UI component
8. Added environment variable in docker file for reading badgr  credential

## Additional info
- We still need to add encrypted variable in cirrus CI for badgr if you want the feature to be ready

- Currently the feature only can issuing badge from badgr credential provided in environment variable. But later on, we can change that


avowkind:
Need to increase test coverage to pass the Codecov tests.

components/IssueBadge/issueBadge.js
```diff
@@ -53,8 +52,7 @@ function BadgeList ({ badgeList, setBadgeChosen }) {
 }
 
 const setAvailableBadgeData = async (setBadge) => {
-  const response = await fetch(`${config.appUrl}/api/badges`)
-  const data = await response.json()
+  const data = await callApi(`badg  es`)
```
avowkind: This doesn't look right badg es  with a space in it?

PR #394, opened 2019-09-22
Vp 301 collect and show the number of volunteers required

aswinks88:
# Pull Request Template (VP-123)

Please check that your ticket number and ticket description is included in the PR title 😀

## Proposed Changes? 🤔
1. Activity providers can now use the newly added input fields to specify the number of volunteers required 
2. Added new entity 'volunteers' in activity.js to store the number of volunteers required in the DB
3. 

## Additional Info.🧐

Any additional info goes here

## Screenshots 📷
 
Capture1


aswinks88:
> Hi Aswin -great work. Some changes required:
> ActDetailForm:53 String needs to be a Formattedmessage for translation.
> 
> You need to add unit tests to ActDetailForm.spec.js as the new actRadio line 47-70 don't have test coverage. The tests should validate toggling the radio buttons enables one or other of the fields, and that you can use either field but not both and that the result returned is either a fraction or an integer.

aswinks88:
> Hi Aswin -great work. Some changes required:
> ActDetailForm:53 String needs to be a Formattedmessage for translation.
> 
> You need to add unit tests to ActDetailForm.spec.js as the new actRadio line 47-70 don't have test coverage. The tests should validate toggling the radio buttons enables one or other of the fields, and that you can use either field but not both and that the result returned is either a fraction or an integer.

Hi Andrew, thanks for the feedback. I will do the necessary changes and update the branch asap.

aswinks88:
> Good progress. More work required.
> Some strings were not in FormattedMessage blocks - i.e were not translatable. I've updated those.
> 
> When the form is opened for editing again the boxes are not initialised with the previous value and neither radio button is enabled.
> 
> If I put 5 volunteers into box 1 this correctly displays on the detail page as 5 volunteers needed but If I put 5 students into box 2 this shows the same thing and should be showing 1 volunteer per 5 students. Also the value stored in both cases is "volunteers": 5, This can't be correct. in case 2 the value stored should be 0.2 (1/5)
> 
> Finally on ActDetail where the value is displayed you will need a test to see if the value is greater or less than 1 if > 1 then show N volunteers required
> if < 1 show 1 volunteer for each N students required
> if = 1 show 1 Volunteer required.
> 
> All strings should be in translation blocks - <FormattedMessage
> 
> Thanks, You are progressing, keep going.

Sure, I am working on it at the moment and will update the branch with necessary changes.

aswinks88:
> > Good progress. More work required.
> > Some strings were not in FormattedMessage blocks - i.e were not translatable. I've updated those.
> > When the form is opened for editing again the boxes are not initialised with the previous value and neither radio button is enabled.
> > If I put 5 volunteers into box 1 this correctly displays on the detail page as 5 volunteers needed but If I put 5 students into box 2 this shows the same thing and should be showing 1 volunteer per 5 students. Also the value stored in both cases is "volunteers": 5, This can't be correct. in case 2 the value stored should be 0.2 (1/5)
> > Finally on ActDetail where the value is displayed you will need a test to see if the value is greater or less than 1 if > 1 then show N volunteers required
> > if < 1 show 1 volunteer for each N students required
> > if = 1 show 1 Volunteer required.
> > All strings should be in translation blocks - <FormattedMessage
> > Thanks, You are progressing, keep going.
> 
> Sure, I am working on it at the moment and will update the branch with necessary changes.

Hi Andrew,
A quick question on this task........
we need 1 volunteer per 5 students which is 1 / 5 (1:5 ratio) = 0.2. Let's say we need 3 volunteers for 15 students then the ratio would be 1:5 = 0.2. So for every volunteer to student ratio, the value is 0.2. And we would be storing this value for the numbers that are divisible by 5. Would this lead to confusion as we will be storing 0.2 in the database?

I have made a workaround for this problem.  
act.volunteerRatio will have the number of students. When we divide this by 5 it will give us the number of volunteers required. 

**const volreq = act.volunteerRatio / 5** // I have taken volreq which gives the number of volunteers required..... I have used Math.ceil(volreq) just in case if the number of students drops below 5
**const value = volreq / act.volunteerRatio
console.log(value )** // this gives the value 0.2
the output would be 3 volunteer(s) required for 15 students

Please let me know if this solution is suitable? 
Thanks

avowkind:
Hi Aswin - this solution is not suitable and not required. 

Normally the activity specifies the whole (integer) number of volunteers - that is irrespective of the number of students. 

However if the ratios matter then its always expressed as the number of students per volunteer. 
So 3 for 15 is 1 in 5   or 1/5 if its 1 volunteer for 10 students its 1/10 or 0.1.  So the result is always a fraction expressed as a decimal 0.x

Please follow the instructions they are correct. No extra storage fields are required. 

The number 5 never comes into the equation. 

Values from 0 to 1 represent the ratio of volunteers to students
Value of 1 means 1 volunteer 
Value greater than 1 means that many volunteers overall.

Only the teacher knows the number of students there will be - the activity designer doesn’t 


A.

PR #401, opened 2019-09-30
VP-57 add getting started info field.

avowkind:
## Proposed Changes? 🤔
1.  Add info.instructions field to organisation text strings
2. add entry rich text editor to the form. 
3. add content to the orgDetails page - getting started tab. 

## Additional Info.🧐
There are now 6 rich text boxes on the organisation form. This may seem intimidating and will need some UX work.  However they make sense.
1. Front About page
2. Getting started information page for people working with the organisation - general instructions. 
3. Pages for followers, joiners, members and people who look at the members page but are not members (outsiders). 

I suggest we setup some examples of well configured organisations to show what each page means.  - Any more content than this and we need to switch to a CMS.

TODO - page appears to be centred. 

Any additional info goes here

## Screenshots 📷
 
image


waltzaround:
Will find the guilty css and fix in a mo

waltzaround:
Re: UX challenge of six inputs - would be a good UX challenge for a junior to try and design something nicer given that most of the setup only needs to occur once - will log it as a challenge and do it if we don't get any responses 😂

PR #406, opened 2019-10-03
VP-561 Fix opportunies always using action.read for security permissions

wseagar:
# VP-561

## Proposed Changes? 🤔
1. Fix security issue where opportunities always using action.read

## Additional Info.🧐

See https://voluntarily.atlassian.net/browse/VP-561 for explanation

### DIFF
--- a/server/middleware/authorize/authorizeRequest.js
+++ b/server/middleware/authorize/authorizeRequest.js
@@ -1,7 +1,7 @@
 const { Action } = require('../../services/abilities/ability.constants')
 
-const defaultConvertRequestToAction = (method) => {
-  switch (method) {
+const defaultConvertRequestToAction = (req) => {
+  switch (req.method) {
     case 'GET':
       return Action.READ
     case 'POST':
@@ -16,7 +16,7 @@ const defaultConvertRequestToAction = (method) => {
 }
 
 const authorizeActions = (subject, convertRequestToAction = defaultConvertRequestToAction) => (req, res, next) => {
-  const action = convertRequestToAction(req.method)
+  const action = convertRequestToAction(req)
   const authorized = req.ability.can(action, subject)
   if (authorized) {
     next()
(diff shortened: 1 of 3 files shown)


server/middleware/authorize/authorizeRequest.js
```diff
@@ -1,7 +1,7 @@
 const { Action } = require('../../services/abilities/ability.constants')
 
-const defaultConvertRequestToAction = (method) => {
```
avowkind: orginally method passed in here

server/middleware/authorize/authorizeRequest.js
```diff
@@ -16,7 +16,7 @@ const defaultConvertRequestToAction = (method) => {
 }
 
 const authorizeActions = (subject, convertRequestToAction = defaultConvertRequestToAction) => (req, res, next) => {
-  const action = convertRequestToAction(req.method)
```
avowkind: and req.method passed in here.  so where is the issue?

wseagar: The issue is in https://github.com/voluntarily/vly2/blob/master/server/api/opportunity/opportunity.routes.js
 with the convertRequestToAction function which has been coded as taking in an req object but is being passed a string.

avowkind: I agree its unclear because the function is called convert request to action so it would be expected to accept a request.   Otherwise, it could be called convertMethodToAction. 

Usually, I would say the function should be a simple as possible and would therefore just use the method string.  However, it's possible that certain actions might need more information from the request to make the decision. In this case, there is value in passing in the req and letting the implementor decide what to use. 

So your changes are correct. the functions take a req.  please update the unit tests to pass in an object { method: 'GET' } as a mock request. 

e.g in AuthoriseRequest.spec.js 66, 70 and 74.

wseagar: Yep sounds good. Thanks for the review.

PR #418, opened 2019-10-11
VP-512 Update email templates and sending logic

avowkind:
## Proposed Changes? 🤔
1.  Interested.controller rationalisation of email sending logic and template naming. - We now give all the interest, person, opportunity and requestor details to the email template.
2.  All new email templates now use the available information for better messages
3.  updated styling to match current website look.
4. expanded test coverage for all email cases. 

## Additional Info.🧐
I investigated switching to react based templates but decided they didn't give us any new functionality of value.  As the pug email templates use blocks the writers only have to concern themselves with overriding key bits and the content is suitable to store int the database. so keep as is for now.

Left to do are the emails for onboarding processes.

## Screenshots 📷
 
image


avowkind:
Note - to test this and see all the templated outputs. change emailperson.js line 23 to 
preview: true,

and then run npm test **/emailperson.spec.js.

This will generate all the emails and load the output into your web browser.

PR #439, opened 2019-10-28
VP-717 latest packages

avowkind:

## Proposed Changes? 🤔
1.  in response to vulnerability warnings in npm install ran npm audit fix, 3 packages updated.
2. Also brought many packages up to the latest version - excluding next 9 and react-intl 4 which have breaking changes
3. as Standard has updated and has many new rules, fixed lint warnings across many files. These are why there are 135 changed files. 
4. notable increase in code coverage 0.7%.
5. warning Browserslist: caniuse-lite is outdated. Please run next command `npm update`  fixed

## Additional information
Particular things now being checked
- use const instead of let if the value is not changed
- no spaces after [ and before ]  i.e. [a, b] not [ a, b ]

- There are a number of new warnings generated about deprecated calls in React 16. 
e.g.
----
Warning: componentWillReceiveProps has been renamed, and is not recommended for use. See https://fb.me/react-unsafe-component-lifecycles for details.

* Move data fetching code or side effects to componentDidUpdate.
* If you're updating state whenever props change, refactor your code to use memoization techniques or move it to static getDerivedStateFromProps. Learn more at: https://fb.me/react-derived-state
* Rename componentWillReceiveProps to UNSAFE_componentWillReceiveProps to suppress this warning in non-strict mode. In React 17.x, only the UNSAFE_ name will work. To rename all deprec⠦ components › Interest › RegisterInterestSection › mount RegisterInterestSection with op and me
----
These originate from Antd.  They will be fixed in AntD 4.0 which I hope won't be too long coming.

## Recommendations
- when updating to this version you will need to run `npm install`. I actually recommend 
`rm -rf node_modules` so that you get all new updates to the libraries.


components/Act/ActDetailForm.js
```diff
@@ -39,7 +39,7 @@ class ActDetailForm extends Component {
     }
 
     this.setImgUrl = this.setImgUrl.bind(this)
-    this.change = this.change.bind(this)
+    this.handleChange = this.handleChange.bind(this)
```
MinhPhu0304: Hi Andrew, the handle change function use arrow function so I think we don't need to bind the function in the constructor anymore.

components/UploadComponent/ImageUploadComponent.js
```diff
@@ -32,7 +32,7 @@ class ImageUpload extends Component {
 
   onUpload (fileIDs) {
     var file = this.uppy.getFile(fileIDs[0])
```
MinhPhu0304: I think var should not be used at all

pages/dbfactory/dbfactory.js
```diff
@@ -9,14 +9,16 @@ export class DBfactory extends Component {
   constructor (props) {
     super(props)
     this.state = { people: 1, ops: 1 }
-    this.queryHandler = this.queryHandler.bind(this)
+    this.handleQuery = this.handleQuery.bind(this)
   }
-  queryHandler = (event) => {
+
+  handleQuery = (event) => {
```
MinhPhu0304: Hi Andrew, handleQuery is a fat arrow function so no need to bind function anymore

PR #471, opened 2019-11-27
VP-822 Fix landing page hero image notch overlap

guttmann:

## Proposed Changes? 🤔
1. At certain screen sizes the hero image "notch" was sitting on top of the navigation bar and in some cases this would cover the search field or the Voluntarily logo. This change updates some of the CSS to ensure the notch goes underneath the navigation bar.

### Original

Example of the problem (blue background is temporary to highlight the notch):

image

### Updated

After fix (blue background is temporary to highlight the notch):

image


guttmann:
@avowkind I was having issues with the testcafe tests passing (that seemed to related to the `wait-on` command timing out). So I have increased the timeout from 30 seconds to 60 seconds which seems to have the testcafe tests passing again now.

I checked this in my local development environment and found that the `wait-on` was taking about 24 seconds to complete when starting up the application so thought this would be a reasonable change. If not, let me know and I can take it out again (or if you'd rather I make that change as part of a separate pull request I can do that too)

PR #472, opened 2019-11-27
VP-823 Set value field of the InputNumber component

bernesemountaindog:
- [ ] Run `npm test` and all the tests pass.
- [ ] Run `npm run lint` and there are no warnings.
- [ ] Included the Jira ID and description in the PR title

## Proposed Changes? 🤔
1. There is a bug because the value field of the InputNumber component wasn't being set. Users could use this component to physically type into, but it wouldn't re-load values after page reload.
2. 
3. 

## Additional Info.🧐

Any additional info goes here

### DIFF
--- a/components/VTheme/NumericRange.js
+++ b/components/VTheme/NumericRange.js
@@ -21,7 +21,9 @@ const Container = styled.div`
 // refs and the new version of reactjs
 class NumericRange extends React.Component {
   render () {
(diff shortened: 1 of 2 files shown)


avowkind:
Voluntari.ly code base branch policy is documented here:  https://trunkbaseddevelopment.com/ <https://trunkbaseddevelopment.com/>. Before deciding to change this you’l need to come up with a strong justification. 

Regards Andrew

PR #492, opened 2019-12-03
VP-855 Include all files in code coverage report

guttmann:

## Proposed Changes? 🤔

1. Configure `nyc` to include all files in code coverage report. This will help identify files that have no tests at all and give a more accurate representation of code coverage.

## Additional Info.🧐

Unfortunately this change will significantly decrease code coverage. (94% -> ~74%)


guttmann:
You've made a few points that I'd like to address

> there are many files in the project we don't want to have to write tests for. - e.g pages/tests components/examples. /x scripts etc

With the "all" configuration turned on `nyc` will still respect the "include" and "exclude" settings. Unit tests, examples, and scripts in the x directory are still not included in the coverage report.

Pages currently are included in the report because we have the rule "pages/**/*.js" in the "include" settings.

> There should be tests for all the top level pages but these are not really unit/functional tests they should be covered by the end to end tests - which we don't really have yet

If a page cannot be tested with unit/functional tests and a decision is made that it will have an end-to-end test instead then we can add that page to the "exclude" list so it is not included in coverage reports.

> identify files that should but don't have unit/functional tests and file a Jira to get them fixed

I'm all for raising Jira cards to add missing tests but from my perspective the best way to keep the missing tests front of mind is to make sure they are in the coverage report.

The other benefit of keeping them in the report is that it will be clear when people add new features that have no tests because coverage will decline and a decision can be made at that point if tests should be added immediately or if a card in Jira can be created to add them at a later date.

> I'm also aware that /pages/api files have unit tests but these don't appear to show in the coverage pages.

I can see the tests in `/__tests__/api` but can also see that they don't appear to affect coverage (e.g. running `/__tests__/api/health/health.spec.js` doesn't give any coverage in the report for `/pages/api/health/*`). It looks like nyc is losing track of things due to the `request` function being used in the tests. I've had a brief look through the issues on the supertest Github repo and there are differing reports of code coverage working/not working when using it. e.g. https://github.com/visionmedia/supertest/issues/213

avowkind:
Thanks for the constructive comments.  I'll merge this in and we can start updating the exclusions.

PR #515, opened 2019-12-08
VP-882/refactor_footer: changed divs and a tags to ul li

careykwan:

## Proposed Changes? 🤔
1. changed footer links to a ul
2. 
3. 

## Additional Info.🧐

Any additional info goes here


bernesemountaindog:
This has caused a regression in the footer. Items are no longer `a` links.

bernesemountaindog:
We'll need to raise a new card to change all the `<a href="..."></a>` tags to instead use `<Link />` component.

See: https://nextjs.org/learn/basics/navigate-between-pages

There are a few examples of this around the website, which is why you sometimes see the app reloading completely instead of acting like a single page app.

bernesemountaindog:
I think we should remove the margin/padding the new `ul` has introduced. As the 'columns' are now indented - not so much an issue in desktop, but in mobile with the limited space it becomes noticeable.

careykwan:
> I think we should remove the margin/padding the new `ul` has introduced. As the 'columns' are now indented - not so much an issue in desktop, but in mobile with the limited space it becomes noticeable.

No worries I will do this now, however im not sure what is cause the test to fail?

PR #521, opened 2019-12-10
Vp 592/redesign organisation page

careykwan:

## Proposed Changes? 🤔
1. New layout for the organisation page "About" tab section
2. 
3. 

## Additional Info.🧐

Any additional info goes here


components/Member/RegisterMemberItem.js
```diff
@@ -131,7 +131,6 @@ class RegisterMemberItem extends Component {
           {options.btns.map((btn, index) => {
             return (
               <span key={index}>
-                {btn.msg && <p>{btn.msg}</p>}
```
avowkind: I don't think the button message should be removed - it's not part of the story and has important functionality. 
If it should be removed you have not removed the btn.msg data elsewhere in the file. So are suppressing output but leaving all the remaining code in. 

image

components/Org/OrgDetail.js
```diff
@@ -1,93 +1,137 @@
...
+
+      @media screen and (min-width: 767px) {
+        font-size: 1rem;;
+      }
+
+      @media screen and (min-width: 988px) {
+        font-size: 2rem;;
+      }
+    }
   }
```
avowkind: This li style looks the same as used elsewhere.  Consider refactoring into a styled component in VTheme. 

If we create new styles in every component the site will not have a consistent look and feel. Also there will be many places that need changing if the theme changes.

components/Org/OrgDetail.js
```diff
@@ -1,93 +1,137 @@
...
-    <Icon type='info-circle' />
+  <span style={{ fontSize: '1rem', fontWeight: 'bold', color: '#6549AA' }}>
     <FormattedMessage id='orgAbout' />
   </span>
 )
 
 const orgMemberTab = (
-  <span style={{ fontSize: '1.2rem', fontWeight: 'bold' }}>
-    <Icon type='team' />
+  <span style={{ fontSize: '1rem', fontWeight: 'bold', color: '#6549AA' }}>
```
avowkind: Same here - two fragments using the same style - if we want a bold purple label then we should put it in the theme or base css class so that it can be used consistent.y

components/Org/OrgDetail.js
```diff
@@ -107,70 +150,60 @@ const orgInstructionTab = (
   </span>
 )
 
+const orgOffersTab = (
+  <span style={{ fontSize: '1rem', fontWeight: 'bold', color: '#6549AA' }}>
```
avowkind: replace this span with a styled component.

components/Org/OrgDetail.js
```diff
@@ -107,70 +150,60 @@ const orgInstructionTab = (
...
+  <OrgPageWrapper>
     <Head>
       <title>Voluntarily - {org.name}</title>
     </Head>
     <PageHeaderContainer />
     <ProfileContainer>
       <ProfileImage src={org.imgUrl} alt={org.name} />
       <TitleContainer>
-        <H3Bold>{org.name}</H3Bold>
+        <OrgHeader>{org.name}</OrgHeader>
```
avowkind: How will OrgHeader be kept looking the same as the Activity and Opportunity Page headers?  All page headers should be the same hence the use of a common styled component.  
Consider refactoring.

components/Org/OrgDetail.js
```diff
@@ -107,70 +150,60 @@ const orgInstructionTab = (
...
-            </ButtonGroup>
+            <h2>About</h2>
+            <Markdown style={{ fontSize: '1.5rem' }} children={(org.info && org.info.about) || ''} />
+          </TabContainer>
+          <Divider />
+          <TabContainer>
+            <h2>Contact</h2>
+            <ContactList>
+              <li>Website<br /><a href={org.website} target='_blank' rel='noopener noreferrer'>{org.website}</a></li>
+              <li>Twitter<br /><a href={org.twitter} target='_blank' rel='noopener noreferrer'>{org.twitter}</a></li>
```
avowkind: You have changed the behavior of the page so that the word Twitter and Facebook is shown even when the org does not have these links.   The previous code was correct and you do not need to change it.

components/Org/OrgDetail.js
```diff
@@ -1,93 +1,137 @@
...
+import { PageHeaderContainer } from '../VTheme/VTheme'
+import OrgDetailForm from '../Org/OrgDetailForm'
 
 function callback (key) {
   // TODO: [VP-300] on tab change update the path so that the page is bookmark and reloadable
 }
 const shadowStyle = { overflow: 'visible', textAlign: 'center' }
 const { TabPane } = Tabs
 
+const OrgPageWrapper = styled.div`
```
avowkind: Please remove this and apply the changes to FullPage.  There is no reason for OrgPage to behave differently to all the other pages.  Each page style should be defined only once and used everywhere, otherwise, we will waste a lot of time trying to get the pages consistent. 
Please Move these style elements to VTheme and agree on them with Walt before implementing changes to how a page looks.
I think a good rule is that there are no @media styles outside of VTheme components.  Page behaviour, Title styles and section wrappers MUST be consistent everywhere. 
So no local styles except for unusually placed components.

components/Member/RegisterMemberItem.js
careykwan: The button message was removed in the new designs that @waltzaround  provided, please discuss with @waltzaround if you wish to leave this in the new design and let me know and I can change this back.

components/Org/OrgDetail.js
careykwan: I have moved this styled ul to the VTheme page, unsure if this new styled list will actually be used elsewhere other then the organisation page.

components/Org/OrgDetail.js
careykwan: Made a component called OrgTabs and moved to Vtheme.

components/Org/OrgDetail.js
careykwan: Currently the Activity and opportunity page headers are h1, I have now deleted OrgHeader and changed it to a h1 so its the same as the activity and opportunity page.

components/Org/OrgDetail.js
careykwan: Sorry I have changed this back now.

components/Org/OrgDetail.js
careykwan: I have changed this back to FullPage, I initially made my own component as I wanted to removed the large margin at the bottom of the FullPage component but did not want to make site wide changes (so I didn't want to touch FullyPage in Vtheme) and I only wanted it removed on the Orgpage for the time being. 

@media screen styles are used throughout the pages already. This is a good way to keep styled components responsive.

PR #522, opened 2019-12-11
VP-877 School invite magic button link handler

guttmann:

## Proposed Changes? 🤔
1. Makes the button in the school invite email actually do something. Clicking the button will prompt the person to sign in/up, create an organisation from their school, make them an org admin, and send them to their dashboard to continue the rest of their journey.

## Additional Info.🧐

While working on this feature I noticed that the org detail form had a form field for "Contact email" but the schema for organisations didn't have this field defined so I've defined it as part of this work (so I could copy across the contact email we have for the school).

I also updated the import school script to import the contact name and email for the school as these weren't previously being imported.

I was also getting errors with the import/export being used in `pages/api/token/token.js` and had to update that file to use require/module.exports instead.

This is the final sub task for VP-826, we should be able to mark that card as complete once this is approved/merged.


server/api/school-invite/school-invite.controller.js
```diff
@@ -124,6 +131,52 @@ class SchoolInvite {
       return false
     }
   }
+
+  static async accept (request, response) {
+    return handleToken(request, response, {
+      join: async (props) => {
+        const organisation = await SchoolInvite.createOrganisationFromSchool(props.schoolId)
+        await SchoolInvite.linkPersonToOrganisationAsAdmin(organisation._id, request.session.me._id)
+      }
```
guttmann: @avowkind from our conversation last week - I believe this is where you'll want to add your code to create goal cards for this person

avowkind: Thanks - I'll be able to issue a pub/sub notice here and have it handled in a more general location.

server/api/school-invite/school-invite.controller.js
```diff
@@ -124,6 +131,52 @@ class SchoolInvite {
...
+    })
+  }
+
+  static async createOrganisationFromSchool (schoolId) {
+    const schoolData = await SchoolLookUp.findOne({ schoolId: schoolId })
```
avowkind: perhaps should check here that you got a result. I presume schoolId originated from the db lookup so should not fail.
However the db could have other issues so you'll want to check

guttmann: Yep makes sense - I'll add that now

PR #547, opened 2019-12-18
VP-945 Fix loadGoals test occasionally failing

guttmann:

## Proposed Changes? 🤔

1. The loadGoals test isn't `await`ing the promise returned by the `loadGoals` function so sometimes the test will fail because not all goal records are loading in time.

### DIFF
--- a/server/api/goal/__tests__/loadGoals.spec.js
+++ b/server/api/goal/__tests__/loadGoals.spec.js
@@ -19,7 +19,7 @@ test.after.always(async (t) => {
 test('Should correctly give count of all goals sorted by rank', async t => {
   let count = await Goal.countDocuments()
   t.is(count, goals.length)
-  loadGoals()
+  await loadGoals()
   count = await Goal.countDocuments()
   t.is(count, 10) // union of the two sources
 })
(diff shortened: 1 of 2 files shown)


bernesemountaindog:
This lint package can be quite useful: https://www.npmjs.com/package/eslint-plugin-promise
It'll highlight promises which are left dangling

avowkind:
> This lint package can be quite useful: https://www.npmjs.com/package/eslint-plugin-promise
> It'll highlight promises which are left dangling

It would be nice to have - but will it work alongside jsstandard?  we don't want two competing lints.

guttmann:
> > This lint package can be quite useful: https://www.npmjs.com/package/eslint-plugin-promise
> > It'll highlight promises which are left dangling
> 
> It would be nice to have - but will it work alongside jsstandard? we don't want two competing lints.

I've had a quick look - it doesn't look like we can do it while using `standard` itself (i.e. running the `standard` command for linting). However, there are a couple of ways we could achieve it without maintaining two competing linting tools.

1. Switch over to using `standardx` (https://github.com/standard/standardx) which allows customisation of eslint plugins and rules
2. Switch to `eslint` using the standard config (https://www.npmjs.com/package/eslint-config-standard) which would then also allow us to customise plugins and add additional rules

It looks like the move to `standardx` would be simpler (and there is a migration tool available too: https://github.com/josephfrazier/standard-eject).

It's really up to you though @avowkind if you think the benefits of being able to add additional rules (like the promise checks) is worth the effort of switching over.

avowkind:
Its a possibility - but maybe for another time.

PR #548, opened 2019-12-18
Vp 864/school address on activity page

careykwan:

## Proposed Changes? 🤔
1. Added venue field to activity form
2. Venue address is show on the activity card
3. The address can be clicked on and be taken to the google maps

## Additional Info.🧐
@avowkind I had trouble working out how to populate the org address to the venue field so I will make another card to add this in later, but in the card, the venue field has been added and the address is show in the op detail page and can be clicked on to be viewed in google maps.

Any additional info goes here

### Updated
image

image

image


server/api/opportunity/opportunity.js
```diff
@@ -30,7 +30,9 @@ const opportunitySchema = new Schema({
   requestor: { type: Schema.Types.ObjectId, ref: 'Person', required: true },
   dateAdded: { type: 'Date', default: Date.now, required: true },
   href: String,
-  tags: { type: [String], required: true, default: [] }
+  tags: { type: [String], required: true, default: [] },
+  organisationAddress: String,
```
avowkind: This should not be the organisationAddress.  If present it should be the address where the activity is taking place.  i.e locationAddress.  It should be initialised from the op.offer Org organisation address.

server/api/opportunity/opportunity.js
```diff
@@ -30,7 +30,9 @@ const opportunitySchema = new Schema({
...
+  organisationName: String
```
avowkind: This is not required-  The organisation name is already present in the offerOrg field.

components/Op/OpDetailForm.js
```diff
@@ -353,6 +385,36 @@ class OpDetailForm extends Component {
...
+                    ]
+                  }
+                  )(<Input placeholder='Name' />)}
+                </Form.Item>
+                <Form.Item
+                  label={opAddress}
+                  validateStatus={organisationAddressError ? 'error' : ''}
+                  help={organisationAddressError || ''}
+                >
+                  {getFieldDecorator('organisationAddress', {
```
avowkind: An address is not mandatory. 
Lots of activities may take place at non specific locations - e.g ONLINE.

If the location select is set to Online - then the address input should not be shown at all.

components/Op/OpDetailForm.js
```diff
@@ -166,6 +168,34 @@ class OpDetailForm extends Component {
...
+          <Icon type='question-circle-o' />
+        </Tooltip>
+      </span>
+    )
+    const opAddress = (
+      <span>
+        {' '}
+        <FormattedMessage
+          id='opAddress'
+          defaultMessage='School or organisation address'
```
avowkind: Not the school address.  The place where the event will be held.

PR #571, opened 2020-01-09
VP-969 Add tests for NoResult component

guttmann:

## Proposed Changes? 🤔
1. Add tests for NoResult component


guttmann:
Hmm, given the error during the build this may take more rework than expected to get working properly.

The error:

```
Failed to compile.
./components/NoResult/NoResult.js
SyntaxError: /usr/src/app/components/NoResult/NoResult.js: [React Intl] Messages must be statically evaluate-able for extraction.
  16 |     return (
  17 |       <FormattedMessage
> 18 |         id={id}
     |             ^^
  19 |         defaultMessage={msg}
  20 |         description={description}
  21 |       />
```

I was wondering if this would become a problem - possible fix could be to move the `FormattedMessage` out of the `NoResult` component and pass that in as a prop when the `NoResult` component is used.

The only current usage of the `NoResult` component is here:

https://github.com/voluntarily/vly2/blob/0d6358ee62a2f5b559958a5c1d4c45371def3880/pages/act/actlistpage.js#L93-L97

With the proposed fix this code would become something like this:

```
<NoResult
  title={
    <FormattedMessage
      id='act.noresult'
      msg='No activities found based on your search criteria'
      description='Message shown while no activities found'
    />
  }
/>
```

avowkind:
The whole component is a bit iffy.  
It’s really trying to be an extension of a formatted message with an icon which has a default. 

Better syntax would be to use {children} property
<NoResult>
 <FormattedMessage
      id=‘act.noresult'
      msg='No activities found based on your search criteria’
      description='Message shown while no activities found’
  />
</NoResult>

<NoResult />. With no children can then display a fixed generic message

guttmann:
I agree that using the children to pass in the `FormattedMessage` looks cleaner, the main reason I was thinking of using the `title` prop is because that mimics the API of the `Result` component (which the `NoResult` component wraps)

guttmann:
I'm going to close this for now and come back to it

PR #594, opened 2020-01-23
VP-333 Always show interested status when navigating away

RyleeQI:

## Proposed Changes? 🤔
1. Will always show the interested status after navigating away


components/Interest/RegisterInterestSection.js
```diff
@@ -24,10 +25,10 @@ class RegisterInterestSection extends Component {
   // When component mounts, make initial API call.
   // TODO do we need to change this to getInitialProps?
   async componentDidMount () {
-    const op = this.props.op
+    const op = this.props.op._id
     const me = this.props.meID
     try {
-      await this.props.dispatch(reduxApi.actions.interests.get({ id: '', op, me }))
+      await this.props.dispatch(reduxApi.actions.interests.get({ op, me, z: cuid() }))
```
avowkind: We need to show why we are adding the z: cuid() query parameter. Otherwise someone might think it is not being used and remove it.  Probably renaming z to cacheBreak: cuid()  would be sufficient.

components/Interest/__tests__/RegisterInterestSection.spec.js
```diff
@@ -59,7 +59,8 @@ test.serial('mount RegisterInterestSection with with no existing interest', asyn
   const realStore = makeStore(initStore)
   const myMock = fetchMock.sandbox()
   reduxApi.use('fetch', adapterFetch(myMock))
-  const getmyinterests = `${API_URL}/interests/?op=${opid}&me=${meid}`
+  // const getmyinterests = `${API_URL}/interests/?op=${opid}&me=${meid}`
```
avowkind: remove unused commented out code

server/api/story/__tests__/story.spec.js
```diff
@@ -51,5 +51,5 @@ test.serial('Should correctly give count of all stories sorted by name', async t
     .expect('Content-Type', /json/)
   const got = res.body
   t.is(t.context.stories.length, got.length)
-  t.is(got[0].name, stories[0].name)
+  t.is(got.name, stories.name) // TODO: have to sort with datePublished to prove the oldest stories
```
avowkind: This is a good example of a test that cannot fail. 
If got is an array then got.name is undefined.  ditto stories.name so the tests always passes because you compared two undefined values. 

Use node to prove to your self that .value of an array doesn't throw an error and just returns undefined.

server/api/story/story.controller.js
```diff
@@ -3,7 +3,7 @@ const Story = require('./story')
 const getStory = async (req, res) => {
   try {
     const got = await Story.findOne(req.params)
-      .populate('author', 'name body imgUrl')
+      .populate('author', 'name body imgUrl') // TODO: add test that will check name and imgURL are populated
```
avowkind: if you add a TODO: in the code you must do one of the following:
a) DO the todo and remove it
b) create a Jira for the TODO so that it gets done in the future. 
c) decide it doesn't need doing and remove the TODO.

components/Interest/RegisterInterestSection.js
RyleeQI: Ok! I will add some comments on it.

components/Interest/__tests__/RegisterInterestSection.spec.js
RyleeQI: OK!

server/api/story/__tests__/story.spec.js
RyleeQI: Hi Andrew. Could you explain what should I do for this comment?

RyleeQI: I'm a little bit can't understand

RyleeQI: @avowkind

PR #595, opened 2020-01-23
Vp1050 fix op page layout

waltzaround:
- [ x] Run `npm test` and all the tests pass.
- [ x] Run `npm run lint` and there are no warnings.
- [ x] Not decreased the overall test coverage?
- [ x] Included the Jira ID and description in the PR title

## Proposed Changes? 🤔
1. Fix OpAbout layout
2. Move sharelinks to OpAbout
3. Small tweak to OpManage to improve consistency - refer to screencap
4. Improve OpTab styling

## Screenshots 📷
 Screen Shot 2020-01-23 at 5 36 51 PM
Screen Shot 2020-01-23 at 5 37 02 PM


waltzaround:
This is quite weird - tests ran fine on local repeatedly, but failed twice on cirrus before passing on the third attempt

avowkind:
I've a slight concern that the Danger Zone is a little too scary and might cause OPs to not Complete their Event.  Can we call it something like 'Careful Now'

components/Op/OpAboutPanel.js
```diff
@@ -5,24 +5,43 @@ import TagDisplay from '../Tags/TagDisplay'
...
+            {description}
+          </Html>
+          <Divider />
+          <TagContainer>
+            <h5><FormattedMessage id='actCategories' /></h5>
+            <TagDisplay tags={op.tags} />
+          </TagContainer>
+
+          <Divider />
+          <h5><FormattedMessage id='actShare' /></h5>
```
avowkind: FormattedMessage MUST have a defaultMessage.

This is a problem as it uses an id without providing a defaultMessage or description. Only Adding it to en.json will cause a break when the language file is regenerated.

components/Op/OpAboutPanel.js
```diff
@@ -5,24 +5,43 @@ import TagDisplay from '../Tags/TagDisplay'
...
+          <ShareLinks url={appUrl} />
+        </div>
+      </OpSectionGrid>
+      <Divider />
+      <OpSectionGrid>
+        <h2><FormattedMessage id='actDetailForm.AboutSection.organisersubtitle' /></h2>
```
avowkind: As above - must have a defaultMessage

components/Op/OpTabs.js
```diff
@@ -6,7 +6,7 @@ import { OpAboutPanel } from './OpAboutPanel'
 import { OpQuestionPanel } from './OpQuestionPanel'
 import { OpUpdatePanel } from './OpUpdatePanel'
 import { OpManagePanel } from './OpManagePanel'
-const shadowStyle = { overflow: 'visible', textAlign: 'center' }
+const shadowStyle = { overflow: 'visible', textAlign: 'center', fontWeight: 'bold', color: '#6549AA' }
```
avowkind: Can we make tabs the same style across the platform by creating VTabs in VTheme?

components/Op/OpAboutPanel.js
avowkind: Do the share buttons really need a title?

waltzaround:
Activity settings?

components/Op/OpAboutPanel.js
waltzaround: I would say so - it fits the format for secondary sections imo

ie: categories get a title

waltzaround:
Finish activity?

components/Op/OpTabs.js
waltzaround: Will put in a ticket and separate PR for that

components/Op/OpAboutPanel.js
waltzaround: Fixed in latest commit

components/Op/OpBanner.js
```diff
@@ -50,7 +49,6 @@ const OpBanner = ({ op, children }) => {
           <h1>{op.name}</h1>
           <ItemIdLine item={op.offerOrg} path='orgs' />
           <h4>
-            {op.subtitle} <br />
             {op.location}
```
avowkind: I disagree that subtitle should not appear on the OpDetail page.
My reason is that the page is the target for shares on social media (not the OpCard). 
1. we want the shared presentation to look like the opCard as much as possible.  so the subtitle should be in the metadata and 'content' for the page for Search Engine results. 
2. Its likely volunteers will be directly straight to this page bypassing the OpCard.  The Subtitle is the Call to action for the volunteer - so needs to be prominent on the banner and definitely in the metadata.

If not present in the banner it should move to the top of the About panel and be the subheading there.

waltzaround: 1. agree that it should be in meta
2. happy to test it out but I'm suspecting it'll result in a crapload of double-ups where organisers explain themselves twice later on

waltzaround: I've chucked it in for now though

PR #596, opened 2020-01-23
VP-1040 Add sendEmailNotifications flag to person

guttmann:

## Proposed Changes? 🤔
1. Add a new field to the person schema to track whether or not a person wants to receive email notifications from Voluntarily. The field is a required boolean field that defaults to true (opted in to notifications).

## Additional Info.🧐

This flag will be used to decide if we send the user email notifications. VP-1041 will add the ability to update this flag on a person's profile, VP-1042 will update any existing emails to respect this flag when deciding whether or not to send an email notification

### DIFF
--- a/server/api/person/__tests__/person.spec.js
+++ b/server/api/person/__tests__/person.spec.js
@@ -264,3 +264,28 @@ test('Should correctly handle missing inputs', async t => {
   t.is(res.body.message, 'Person validation failed: email: Path `email` is required.')
   t.is(res.body.name, 'ValidationError')
 })
+
+test.serial('Email notifications flag set correctly', async (t) => {
+  const people = await Person.create([
+    {
+      name: 'Test default flag',
+      email: 'test1@example.com'
+    },
+    {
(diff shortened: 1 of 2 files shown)


guttmann:
> Noting that the new field has not been added to the Ability constants and thus won't be sent out in API requests for a person - this is ok as the test flag is used client side. However, we will need to check what happens if someone modifies their profile and updates with a change record with no value for sendEmailNotifications - would it stay the same or return to the default i.e turn itself back on.
> 
> Could you check that editing one's profile does not re-enable the flag.

Yep I have just checked that, I took the following steps:

1. Made sure my account's role was just "volunteer"
2. Turn off email notifications flag and save
3. Update another field on my profile and save
4. Confirm that email notifications flag was still disabled

Also did the opposite where I turned it on, modified another field and saved, and checked the flag was still on.

Looking at the `PUT` requests each time I hit save - all of my profile data is included in every request regardless of which field I updated. Is there a scenario where that won't be the case?

PR #597, opened 2020-01-23
Vp 929 - when ticking 'school' as the organisation type, hid extra fields

careykwan:
- [ ] Run `npm test` and all the tests pass.
- [ ] Run `npm run lint` and there are no warnings.
- [ ] Included the Jira ID and description in the PR title

## Proposed Changes? 🤔
1. when ticking 'school' as the organisation type, un-necessary fields at the bottom are hidden.
2. 
3. 

## Additional Info.🧐

Any additional info goes here

### Original
image

### Updated
image

### DIFF
--- a/components/Org/OrgDetailForm.js
+++ b/components/Org/OrgDetailForm.js
@@ -448,115 +448,120 @@ class OrgDetailForm extends Component {
                 <Divider />
               </>)
             : null}
-          <FormGrid>
-            <DescriptionContainer>
-              <TitleContainer>
-                <h3>Staff Instructions</h3>
-              </TitleContainer>
-              <p>
-                What processes do you want staff to do before they start
-                volunteering? Should they fill in a form? Ask for approval from
-                their manager? Specify your instructions here.
-              </p>
-            </DescriptionContainer>
-            <InputContainer>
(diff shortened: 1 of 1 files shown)


avowkind:
Change looks good 

Note that removing the ability to add content to the info fields means that there is nothing to display on the 'members' tab - unless you are orgAdmin.  So perhaps the tab should be hidden except for admins. 

Are you sure there is nothing a school would want to say to its followers?

careykwan:
> Change looks good
> 
> Note that removing the ability to add content to the info fields means that there is nothing to display on the 'members' tab - unless you are orgAdmin. So perhaps the tab should be hidden except for admins.
> 
> Are you sure there is nothing a school would want to say to its followers?



> Change looks good
> 
> Note that removing the ability to add content to the info fields means that there is nothing to display on the 'members' tab - unless you are orgAdmin. So perhaps the tab should be hidden except for admins.
> 
> Are you sure there is nothing a school would want to say to its followers?

I just asked Sam about this and think we are going to review the content first and see if we actually want to hide the fields or change the fields so they are more relevant to teachers. I will just decline this pull request for now.

PR #612, opened 2020-01-29
VP-1077 - ensure production public key is in the release.

avowkind:

## Proposed Changes? 🤔
In production mode the public and private keys for token generation were being read from environment variables. However only the private key was being set in the container. This meant that jwt verify fails because it used a different pub key to verify (a warning string).  This resulted in the invalid algorithm message.

This change puts the public key prod.pub.pem in the lib/sec folder and reads it into the keys. 

also changed the response from 500 to 404. In this error when we had the wrong keys 500 is the appropriate response however usually the JWT failure would arise from an invalid token and thus we refuse the request - we can't find the resource the jwt is requesting. 

## Additional Info.🧐
Note - public keys are not secret - they are made available so that the JWT can be verified. Only the private key should be encrypted into the container.

### DIFF
--- a/components/Header/Header.js
+++ b/components/Header/Header.js
@@ -85,7 +85,6 @@ const getAllowedLinks = isAuthenticated =>
 const Header = ({ isAuthenticated, ...props }) => {
   const intl = useIntl()
   let notice = intl.formatMessage({ id: 'notice', defaultMessage: 'none' })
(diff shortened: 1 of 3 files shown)


avowkind:
Note - This has to be pushed to alpha to be tested as the fault is dependent on the prod build docker container.

PR #622, opened 2020-02-03
VP-1083 Implement activity API permissions

guttmann:

## Proposed Changes? 🤔
1. Added ability file for Activity API
2. Added some additional checks for access in `activity.controller.js`

## Additional Info.🧐

Permissions implemented as outlined in the API Access Security Rules documentation.

[API Access Security Rules](https://voluntarily.atlassian.net/wiki/spaces/VP/pages/18677761/API+Access+Security+Rules#APIAccessSecurityRules-Activities)


guttmann:
The latest setSession changes appear to be affecting the permissions (which is causing two tests to fail) - it looks like the test user is having the `opportunityProvider` role added a second time (to the end of the roles array) and this is causing the "Activity" create permissions to look like this:

```
[
  { subject: 'Activity', action: 'create', inverted: true },
  { subject: 'Activity', action: 'create', inverted: true },
  { subject: 'Activity', action: 'create', inverted: true },
  { subject: 'Activity', action: 'create' },
  { subject: 'Activity', action: 'create' },
  { subject: 'Activity', action: 'create', inverted: true }
]
```

The last record there is because of this extra opportunityProvider role and unfortunately it looks like CASL will honour the last rule in the list if there are conflicting rules.

I was able to create a simple reproduction of this issue on runkit with this code:

```
var ability = require("@casl/ability")

const test = new ability.Ability([
  { subject: 'Activity', action: 'create', inverted: true },
  { subject: 'Activity', action: 'create', inverted: true },
  { subject: 'Activity', action: 'create', inverted: true },
  { subject: 'Activity', action: 'create' },
  { subject: 'Activity', action: 'create' }
])

console.log(test.can('create', 'Activity')) // returns `true`

const test2 = new ability.Ability([
  { subject: 'Activity', action: 'create', inverted: true },
  { subject: 'Activity', action: 'create', inverted: true },
  { subject: 'Activity', action: 'create', inverted: true },
  { subject: 'Activity', action: 'create' },
  { subject: 'Activity', action: 'create' },
  { subject: 'Activity', action: 'create', inverted: true }
])

console.log(test2.can('create', 'Activity')) // returns false
```

So the order of roles and the resulting rules that get built from them is very important and we probably need a better solution to handle this...

guttmann:
Example reproduction on RunKit: https://runkit.com/embed/pl8f47713045

avowkind:
I can remove the duplication easily enough but what is the expected ordering ?  do we need the most permissions first or last?

avowkind:
What happens in the case where a person has both OP and AP role with AP following OP in the role list? Then the created ability for Activities would deny for the op and enable for the ap.  However, the same order list for the op create ability would enable for op first and then disable for the ap role leaving the permission disabled.  
Seems like the abilities should be additive - starting with deny all and then enabling as required.

guttmann:
Yep that sounds similar to what I was thinking

A quick fix for the time being would be to remove duplication and then ensure the `userRoles` array in `getAbility` is sorted from least permissive to most permissive.

However, we need a bigger change to solve the OP+AP problem you've highlighted.

I wonder if we need the `ruleBuilder` functions to return just the array of rules rather than the map of roles to rules. Then we can make it the responsibility of each `ruleBuilder` function to come up with the correct set of rules for a specific end point (based on the authenticated person's roles). This would hopefully avoid the duplicate/conflicting rules ending up in the ability.

We could make a piece by piece transition to this new approach by updating the `getAbility` function to use the current approach when a `ruleBuilder` returns an object and the new approach where it returns an array - that would allow us to convert each ability file one by one without affecting existing functionality.

What do you think @avowkind?

avowkind:
I've added sortRoles to member.lib.js so the session roles are in a standard order and deduplicated with VP first and Admin last. There's still some failures in the activity.ability.spec.js.

avowkind:
Yes you are on the right track. 

I do have a concern with the way the rules are built.  we seem to have to read all the .ability files and process them for each request.  We get a lot of requests only a few of which need ability rules. i.e only /api files need ability processing.  We cover pages using securePage.  For any given API/endpoint we probably only need to load the rules for that endpoint. e.g. only need person.abilities for /api/people.  as you suggest.

I suspect that the files are dynamically loaded because they contain tests against the current session. However, these tests should be able to be replaced with functions so that the abilities  files only need to be parsed on server.js startup.  

Unfortunately, the CASL stuff is something I delegated so don't know so well, but I can catch up.

PR #629, opened 2020-02-05
VP-1229 catch errors in badge api if server not available

avowkind:

## Proposed Changes? 🤔
1. add try catch block around setAvailableBadgeData to handle error returned if the Badgr server is not available or if the platform has not set access credentials.

error was visible only on person detail page when signed in as admin so that the Add Badge button was present.

### DIFF
--- a/components/IssueBadge/issueBadge.js
+++ b/components/IssueBadge/issueBadge.js
@@ -56,8 +56,12 @@ function BadgeList ({ badgeList, setBadgeChosen }) {
 }
 
 const setAvailableBadgeData = async (setBadgeList) => {
-  const badgeList = await callApi('badges')
-  setBadgeList(badgeList)
+  try {
+    const badgeList = await callApi('badges')
+    setBadgeList(badgeList)
+  } catch (e) {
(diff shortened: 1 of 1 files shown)


components/IssueBadge/issueBadge.js
```diff
@@ -56,8 +56,12 @@ function BadgeList ({ badgeList, setBadgeChosen }) {
 }
 
 const setAvailableBadgeData = async (setBadgeList) => {
-  const badgeList = await callApi('badges')
-  setBadgeList(badgeList)
+  try {
+    const badgeList = await callApi('badges')
+    setBadgeList(badgeList)
+  } catch (e) {
+    console.error('Either Badgr server is down or you have not set BADGR credentials in environment', e)
```
guttmann: What's the impact on the user when this catch is triggered? Will they still be able to use the `IssueBadgeButton` or do we need to display an error message of some kind so that they're aware something is broken?

avowkind: The result will be there are no badges to issue.  An error would be nice but as this is 99% a dev issue the console log should suffice.

guttmann: OK makes sense :+1:

PR #631, opened 2020-02-05
remove imageResize for now

avowkind:
- [ ] Not decreased the overall test coverage? - necessary decrease for hotfix

## Proposed Changes? 🤔
The imageResize function is replaced with a null passthrough and the sharp library is removed from the build. This is because. linux-x64' binaries cannot be used on the 'linuxmusl-x64

## Additional Info.🧐
removing the feature means that uploaded images will not be optimsed for size and the small and large versions will both be large versions.  This increases storage requirements and download time. 
However this just puts us back where we were a week ago.


bernesemountaindog:
Strange that this doesn't work on the server, oh well, I'll see if I can find time to get it working at a later stage

PR #642, opened 2020-02-10
VP-1109 Draft opportunities don't appear in skills section of home page

farjanacan:

## Proposed Changes? 🤔
1. Draft opportunities don't appear in the recommended and skills section of the home page


farjanacan:
> Sorry this is not the correct solution. DRAFT ops should not be sent in response to the request from the API. Otherwise it is possible for someone who is not the owner to view the opportunities before they are published. The solution will involve the recommendation engine only searching published opportunities.
> While doing this check to ensure that the recommendedOps do not include any that are in the past as well.
> 
> Thanks A.

Thanks for the info. I'll re-open the PR, if I can get it working.

PR #645, opened 2020-02-12
add CASL abilities to tags

avowkind:

## Proposed Changes? 🤔
Access to the Tags API is now controlled by Role Based Access Permissions. 
see : https://voluntarily.atlassian.net/wiki/spaces/VP/pages/18677761/API+Access+Security+Rules

Anon people - cannot access tags
Authorised people - can read and list tags but not create or change them
Admins - can perform all CRUD actions.

## Additional Info.🧐
The tags word list collection feature is only partially implemented.  There is currently a single word list (tags) that is used for all tagging activities. It is read by edit forms that want to autocomplete new tag fields.  It is not updated through the API but is updated when a person saves a profile, activity or opportunity containing new tags. 
Admins have full crud access in order to allow the future tidying up of the tag lists.

### DIFF
--- a/server/api/tag/__tests__/tag.spec.js
+++ b/server/api/tag/__tests__/tag.spec.js
@@ -119,9 +119,11 @@ test('Tag API - admin - CRUD', async t => {
 
   // get by id - non admin check
   response = await request(server)
-    .put(`/api/tags/${id}`)
-    .set('Cookie', [`idToken=${jwtData.idToken}`])
+    .get(`/api/tags/${id}`)
+    .set('Cookie', [`idToken=${jwtDataDali.idToken}`])
+
   t.is(response.statusCode, 200)
+  t.is(response.body.tags.length, 3)
 
   // update
   response = await request(server)
--- a/server/api/tag/__tests__/tag.spec.js
+++ b/server/api/tag/__tests__/tag.spec.js
@@ -19,18 +19,11 @@ test.before('before connect to database', async (t) => {
   ]
   t.context.people = await Person.create(people).catch((err) => console.error('Unable to create people', err))
   t.context.tags = await Tag.create(tagCollection).catch((err) => console.error('Unable to create tags', err))
-  console.log(t.context.tags)
 })
 
 test.after.always(async (t) => {
   await Tag.deleteMany()
   await t.context.memMongo.stop()
-})
-
-test('verify fixture database has tags', async t => {
-  const p = await Tag.find()
-  t.is(2, p.length)
-  t.truthy(p[0].tags.includes('robots'))
 })
 
 /* -- Anon person */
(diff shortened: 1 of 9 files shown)


server/middleware/authorize/authorizeRequest.js
```diff
@@ -18,6 +18,7 @@ const defaultConvertRequestToAction = (req) => {
 const authorizeActions = (subject, convertRequestToAction = defaultConvertRequestToAction) => (req, res, next) => {
   const action = convertRequestToAction(req)
   const authorized = req.ability.can(action, subject)
+  // console.log('authorizeActions', subject, action, req.ability, authorized)
```
guttmann: I'm not sure what your position on commented out code is but I'd usually recommend removing it

server/api/tag/__tests__/tag.spec.js
```diff
@@ -4,37 +4,135 @@ import { server, appReady } from '../../../server'
...
+    .set('Cookie', [`idToken=${jwtDataDali.idToken}`])
+
+  t.is(response.statusCode, 403)
+})
+
+/* -- Admin person */
+
+test('Tag API - admin - CRUD', async t => {
+  let response = await request(server)
+    .post('/api/tags')
```
guttmann: Given there's only a single record ever returned when listing/reading tags should we disallow creating new ones for everyone?

server/api/tag/__tests__/tag.spec.js
```diff
@@ -4,37 +4,135 @@ import { server, appReady } from '../../../server'
...
+    .set('Cookie', [`idToken=${jwtData.idToken}`])
+    .send({ tags: ['a', 'b', 'c'] })
+
+  t.is(response.statusCode, 200)
+  const newtags = response.body
+  const id = newtags._id
+
+  // get by id - non admin check
+  response = await request(server)
+    .put(`/api/tags/${id}`)
```
guttmann: Similar question to the create one here - is the client ever going to be updating the list of tags (and actually need this access)? It seems like there is a lot of room for error/bugs here - given there is only one record for all tags I could see it being easy to accidentally overwrite the entire list and lose data

guttmann: Just noticed that based on the comments that this possibly should have been a `.get` call?

server/middleware/authorize/authorizeRequest.js
avowkind: Yes the rule is usually to remove it. However, this log line turns out to be often re-instated during debug of CASL as it shows the current sign in status and roles.

server/api/tag/__tests__/tag.spec.js
avowkind: It would be simpler but we are likely to want to init the tags database with standard keywords at some point and the plan is to have more than one dictionary fairly soon. So admins have CRUD ability.

server/api/tag/__tests__/tag.spec.js
avowkind: Admins will get a tag list management UI sometime.  As tags are a pick list and not referenced in the tag arrays of entities, updating the list does not break references where other words have been used. 
Aim is to populate a large list of standard terms. 

good spot - will update.

PR #650, opened 2020-02-14
VP-1089 Interest API permissions

guttmann:

## Proposed Changes? 🤔

1. Implement interest API permissions
2. Update `defaultConvertRequestToAction` to be more generic and removed custom implementations
3. Allow `ruleBuilder` functions to be `async`
4. Fix some minor issues with the interest action handlers (functions in `interest.controller.js`


guttmann:
> The merge with the personalGoal API changes has resulted in all the interest.spec.js and interest.ability.spec.js:473 tests failing.
> Trying to review the changes there's something I'm ignorant of - can you tell me more about what makes getAbility change from a sync to async call. given that this is passed into express.use how do we know the promise is resolved by the time the ability is tested?

`express.use` is expecting each middleware function it receives to call `next` so the function you pass in can be synchronous or asynchronous as long as it calls `next` at the end.

It looks like the tests started failing because an `await` disappeared during the merging - I've put it back in and the tests are passing again

avowkind:
Thank - yes I saw how the merge dropped the await - my bad. 
However I scanned the code and couldn’t see where any of the ruleBuilder implementations needed to be Async. 

A.

PR #654, opened 2020-02-16
VP-1224. token handler must cope with unverified users

avowkind:

## Proposed Changes? 🤔
1. tokens issued in school and membership invites lead new users to the platform. These route through handleToken which pushes non-signed in users to Auth0, when they return the actions usually complete. However if the person is signed-in with a non verified email address we cannot let them continue and the handler would get stuck in a loop.  
This change provides an /unverified page that informs the person that they need to complete email verification before trying to use the system.  
2. 
3. 

## Screenshots 📷
image


pages/api/token/token.js
```diff
@@ -11,9 +11,16 @@ const handleToken = async (req, res, actionTable) => {
...
   // if user is not authenticated then get them in.
   if (!req.session.isAuthenticated) {
-    return res.redirect(`/auth/sign-thru?redirect=${req.originalUrl}`)
+    if (req.session.user && !req.session.user.email_verified) {
+      // if user has signed in but is invalid then tell them
+      return res.redirect('/unverified')
+    } else {
+      // user has not signed in
+      return res.redirect(`/auth/sign-thru?redirect=${req.originalUrl}`)
+    }
```
guttmann: I understand the need for this change based on how we check if someone is authenticated now - but don't you think this will be a pretty poor experience for a first time user?

In the case of the school invite to get the correct outcome their journey must look something like this now:

1. Click "accept invite" link in invite email
2. Get to Voluntarily, be redirected to Auth0
3. Sign up on Auth0
4. Redirect back to Voluntarily and see unverified message
5. Find verify email from Auth0 in their inbox and verify their email address
6. Find the original invite email from Voluntarily and click the accept invite link again

I think things will really start to fall apart for people around steps 5 & 6

Is there any way we can run the `handleURLToken` logic when a user hasn't verified their email address? That would eliminate step 6 from the journey above and seems pretty critical for any functionality that involves inviting new people to the platform.

waltzaround: Agreed with @guttmann - can pretty much guarantee it'll fall apart if we need both 5 and 6 to happen

avowkind: I would be happy to let people continue with unverified emails and just block them from volunteering.

avowkind: after discussion with Sam this should go through as it is for now - as currently the feature is blocked/broken. I'll add a new story to improve the experience.

PR #657, opened 2020-02-17
VP-1295 make tag list name explicit

avowkind:

## Proposed Changes? 🤔
1. the tag collection can contain one or more lists of words but so far we only used one. This default document is returned by the listTags controller. /api/tags.  
In the future, we may group tags into different documents - one for skills, one for activity types etc. 
This change ensures that each dictionary is named and can be accessed independently. 
for backwards compatibility the original default tag list is called 'default' and does not need a ?name=name on the api request for access.

Follow up work - once deployed the database should be modified to name the original default tag list.


waltzaround:
Having a look now - agreed with the multi taglists

PR #660, opened 2020-02-17
VP-983 bookmarkable tabs and refactor home page

avowkind:

## Proposed Changes? 🤔
1. refactor and use hooks
The /home page is the primary interactive page for volunteers and teachers. the source file is very long.  This change factors out the banner, tabs and tab panels into separate components.
It also moves some of the data fetching to the child component - e.g the fields used by the person editor. This avoids loading data the person is unlikely to use. 

2. bookmarkable tabs
When the tab is changed the URL is updated with ?tab=tabname. This can be bookmarked and placed in the history.  When loaded the page switches directly to that tab. This allows links and goals to route people directly to the appropriate tab. e.g the profile button in the menu. 

## Additional Info.🧐

The class is converted to a function. only the current person is passed as props. child components needing ops and interests etc connect to redux directly using useSelect.

### Original
image

### Updated
image


waltzaround:
Reviewing now

avowkind:
> I'd put discover before active, but we can suss that in another ticket imo
I can put discover first - but would want active to be the default tab.

waltzaround:
Sounds good to me

PR #661, opened 2020-02-18
VP-1099 Person permissions

bernesemountaindog:

## Proposed Changes? 🤔
1. Applied permissions to the Person API. In general; ADMIN and TESTER can do most anything, the record owner has permission to their own data, others have a more limited view, only ADMIN can delete users, only ADMIN can change a users email field, role field is more secure (still needs work though)
2. 
3. 

## Additional Info.🧐

Any additional info goes here


server/api/person/person.controller.js
```diff
@@ -1,20 +1,73 @@
...
+  const personId = req.params._id
+  if (!personId) {
+    return res.status(400).send('Missing person identifier')
+  }
+
+  const isSelf = me._id && personId === me._id.toString()
+
+  Person
+    .accessibleBy(req.ability, Action.READ)
+    .findOne(query)
```
avowkind: rather than pick fields out of the result - why not use the select parameter and only get the fields needed?

server/api/person/person.controller.js
```diff
@@ -1,20 +1,73 @@
...
+    .exec(async (_err, person) => {
+      if (person) { // note if person does not exist middleware will already have 404d the result
+        await getPersonRoles(person)
+      }
+
+      const fields = [
```
avowkind: I think its generally more performant to define consts like this outside of the lambda.

server/api/person/person.controller.js
```diff
@@ -1,20 +1,73 @@
...
+
+      const includePersonalFields = (me.role &&
+        (
+          me.role.includes(Role.ADMIN) ||
+          me.role.includes(Role.TESTER)
+        )) ||
+        isSelf
+
+      if (includePersonalFields) {
+        fields.push(PersonFields.EMAIL, PersonFields.PHONE, PersonFields.EDUCATION, PersonFields.JOB, PersonFields.LOCATION)
```
avowkind: I wonder if it might be clearer to define the public and personal fields once in the person.constants.  we might even move the personal items to a single field object.

components/Person/EditablePersonPanel.js
```diff
@@ -27,10 +27,32 @@ export const EditablePersonPanel = ({ person }) => {
     setEditing(false)
   }
   const handleUpdate = async (person) => {
+    // Only a subset of fields can be updated on the server
+    const personData = {
```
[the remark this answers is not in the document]
bernesemountaindog: Yeah this is a tricky one, it all depends on the philosophy of how to secure fields on an API. I feel if a field is sent to the server which is not permitted to be updated then instead of the server ignoring that field and continuing, I instead return 403 forbidden. As the user has requested to update a field they're not allowed to, so they should be told about this.

In general it does make things a bit less forgiving for frontend developers because they need to be more strict with what they send. No longer can they just throw fields at the server and see what sticks.

An example I thought of where this more strict approach makes sense is; if we only want to update a single field, say `email`, and we `PUT` just that field as our body -
* If we ignore that field on the server (because the user is not allowed to update it), then we will still respond with 200 success because the server successfully updated 0 fields. This gives the impression to the frontend that their request worked and the `email` field was updated
* If we instead return forbidden because the `email` field cannot be updated, then this indicates more clearly that the request didn't do all that was asked for

So, for this specific line of code - when we GET a person from the server, it may contain fields which cannot be updated and are readonly (e.g. `dateAdded`). As these can no longer just be sent to the server, we either need to use a blacklist to trim readonly fields, or a whitelist to only send a specific set... I guess it's debatable which is preferred.

server/api/person/__tests__/person.permissions.spec.js
```diff
@@ -0,0 +1,787 @@
...
+    .set('Cookie', [`idToken=${await createPersonAndGetToken([Role.ADMIN])}`])
+    .expect(200)
+
+  t.is(res.body.length, await Person.countDocuments())
+
+  // All fields should be returned. Spot check a single user we know has data for each key.
+  const andrew = res.body.find(user => user.email === 'andrew@groat.nz')
+  t.truthy(andrew)
+
+  const expectedFields = [
```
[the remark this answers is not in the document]
bernesemountaindog: True true, having a set of fields copy-paste around the place isn't nice. I'll try see if I can reduce this.
I guess as a team we should agree on a philosophy; `whitelist` of allowed fields or `blacklist` of denied fields.

server/api/person/person.controller.js
bernesemountaindog: Good point :) I'll change this

PR #668, opened 2020-02-23
split client / server config to secure secrets

avowkind:

## Proposed Changes? 🤔
We collect environment information and flags into a handy config object that is made available to pages and functions.  This config includes the Mongo database URI, SMTP and other key information which would give access to private resources. 

It is easy for client and server side code to include this file. and even if the client page only uses public data in the config, the presence in the javascript of the full object would make it accessible. 

This change splits the /config/config.js file into two separate files. 
clientConfig provides information often used by client side components such as the AppUrl, port and some settings. 
serverConfig provides the above plus access to database and other service keys originating from the environment that should not be able to leave the server. 

## Additional Info.🧐
Note this is not bulletproof as it is still possible for a developer to accidentally include the wrong file. However the clearer naming will help checking that this does not happen. 

All files should use clientConfig unless they specifically require values provided by the server Config.


config/clientConfig.js
```diff
@@ -0,0 +1,40 @@
...
+        Accept: 'application/json',
+        'Content-Type': 'application/json'
+      }
+    },
+    auth: {
+      AUTH0_CLIENT_ID: 'S4yd4VgZ92NIjhwO3vt4h0Gifb9mXv1k',
+      AUTH0_CLIENT_DOMAIN: 'voluntarily.au.auth0.com'
+    },
+    apiVersion: 'v1',
+    BADGR_API: process.env.BADGR_API || 'https://api.au.badgr.io'
```
guttmann: Does the client need access to the Badgr API? Aren't all requests proxied through our own Badge API?

avowkind: correct - will fix.

PR #671, opened 2020-02-24
VP-1312 - Avoid duplicating the organisation if school-invite used twice.

avowkind:

## Proposed Changes? 🤔
A school-invite email contains a token that when used creates a new organisation based on the school id and adds the person as an admin. 

The handler did not check whether the invite had already been used and the organisation created, as none of the fields is required to be unique the second create succeeds and we get two schools with the same name but different ids.  Thus each user of the token gets their own school.

This change uses the slug field to check whether the school already exists. if so it returns the original school. 

The schema is changed so that two organisations cannot have the same name or slug. attempts to create an organisation with the same name as an existing one, or to edit the name of an organisation to be the same should fail.

### DIFF
--- a/server/api/activity/__tests__/activity.spec.js
+++ b/server/api/activity/__tests__/activity.spec.js
@@ -15,6 +15,9 @@ import acts from './activity.fixture.js'
 test.before('before connect to database', async (t) => {
   t.context.memMongo = new MemoryMongo()
   await t.context.memMongo.start()
+  t.context.people = await Person.create(people).catch((err) => `Unable to create people: ${err}`)
+  t.context.orgs = await Organisation.create(orgs).catch((err) => `Unable to create organisations: ${err}`)
+
   await appReady
 })
 
@@ -24,8 +27,6 @@ test.after.always(async (t) => {
 
 test.beforeEach('connect and add two activity entries', async (t) => {
   // connect each activity to an owner and org
-  t.context.people = await Person.create(people).catch((err) => `Unable to create people: ${err}`)
-  t.context.orgs = await Organisation.create(orgs).catch((err) => `Unable to create organisations: ${err}`)
   acts.map((act, index) => {
     act.owner = t.context.people[index]._id
     act.offerOrg = t.context.orgs[index]._id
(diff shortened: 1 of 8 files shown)


server/api/activity/__tests__/activity.spec.js
```diff
@@ -15,6 +15,9 @@ import acts from './activity.fixture.js'
 test.before('before connect to database', async (t) => {
   t.context.memMongo = new MemoryMongo()
   await t.context.memMongo.start()
+  t.context.people = await Person.create(people).catch((err) => `Unable to create people: ${err}`)
+  t.context.orgs = await Organisation.create(orgs).catch((err) => `Unable to create organisations: ${err}`)
```
guttmann: These `.catch` calls will swallow errors and allow tests to run even though fixtures haven't been loaded.

Removing the `.catch` calls and just letting any exceptions get thrown will give much faster feedback to the person running the tests - which is preferable in my opinion

server/api/opportunity/__tests__/opportunity.spec.js
```diff
@@ -23,6 +23,8 @@ import { jwtData } from '../../../middleware/session/__tests__/setSession.fixtur
 test.before('before connect to database', async (t) => {
   t.context.memMongo = new MemoryMongo()
   await t.context.memMongo.start()
+  t.context.people = await Person.create(people).catch((err) => `Unable to create people: ${err}`)
+  t.context.orgs = await Organisation.create(orgs).catch((err) => `Unable to create orgs: ${err}`)
```
guttmann: Similar to my other comment - these `.catch` calls will swallow errors and allow tests to run without the expected data being in the database

server/api/activity/__tests__/activity.spec.js
avowkind: Agreed - but this pattern is repeated across almost all tests. The only time you'll see the error is when the fixture is invalid and throws a mongo error. That will be printed in the output.  I'll leave for now and add a new ticket.

avowkind: https://voluntarily.atlassian.net/browse/VP-1335

PR #672, opened 2020-02-24
VP-1322 new user cannot save profile

bernesemountaindog:

## Proposed Changes? 🤔
1. After locking down the person API, frontend clients can no longer send the email, dateAdded field in their payload
2. 
3. 

## Additional Info.🧐

Any additional info goes here

### DIFF
--- a/components/Home/HomeTabs.js
+++ b/components/Home/HomeTabs.js
@@ -55,7 +55,7 @@ const homeDiscoverTab = (
   </span>
 )
 
-export const HomeTabs = ({ person, onChange, defaultTab, me }) =>
+export const HomeTabs = ({ person, onChange, defaultTab }) =>
   <VTabs defaultActiveKey={defaultTab} onChange={onChange}>
     <TabPane tab={homeDiscoverTab} key='discover'>
       <ProfilePanel>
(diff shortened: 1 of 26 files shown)


components/Home/HomeTabs.js
```diff
@@ -78,7 +78,7 @@ export const HomeTabs = ({ person, onChange, defaultTab }) =>
     </TabPane>
     <TabPane tab={homeProfileTab} key='profile'>
       <ProfilePanel>
-        <EditablePersonPanel person={person} />
+        <EditablePersonPanel person={person} me={me} />
```
avowkind: it is not necessary to pass 'me' down through multiple levels of properties.  me is in the redux session so it is available everywhere. 
use the hook.
const me = useSelector(state => state.session.me)

bernesemountaindog: Sounds good I'll change this

bernesemountaindog: Looks like useSelector is only for NFC, PersonDetailForm is a class so it doesn't like being there
Anyway to access the redux store directly?

avowkind: export default connect(store => ({  me: store.session.me } )(PersonDetailForm)
should do it.

avowkind: Make the existing default export into a named const and wrap that.

bernesemountaindog: Thanks :) That's worked nicely :) I've only used redux a little bit, it solves this case of piping props down and down nicely :)

PR #674, opened 2020-02-25
VP-1070 initial sign up email

avowkind:

## Proposed Changes? 🤔
Send a Welcome email to a new person when they first sign in. 

When a person signs in for the first time a new account is created.  AddPerson publishes a TOPIC_PERSON__CREATE event which is received by person.subscribe. This generates a new email to the person using the Welcome template.
This, in turn, generates a TOPIC_PERSON__EMAIL_SENT event. 

## Screenshots 📷
 
image

### DIFF
--- a/emails/theme/footer.pug
+++ b/emails/theme/footer.pug
@@ -9,13 +9,11 @@ table(width="100%")
                 td(align="center" valign="middle" style="font-size: 12px; line-height: 18px; color:#aaaaaa;")
                   span(class="appleFooter")
                     block footer
-                      p Voluntarily is an initiative run by the
-                        a( href="http://www.pamfergusson.co.nz/") Pam Fergusson Charitable Trust
+                      p Voluntarily is an initiative run by the 
+                        a( href="http://www.pamfergusson.co.nz/") 
+                          span Pam Fergusson Charitable Trust. 
                         if (to && to.email && to.href)
                           span This email was sent to #{to.email}. You're receiving this email because you expressed interest in Voluntarily Opportunities.
-                          span If you'd rather not see us in your inbox in the future, please set your
-                            a(href=to.href) profile to inactive.
-
-                    if (unsubscribeLink)
-                      p
-                        a(href=unsubscribeLink) Unsubscribe
+                      p If you'd rather not see us in your inbox in the future, 
+                        if (unsubscribeLink)
+                          a(href=unsubscribeLink) Unsubscribe
--- a/emails/welcome/html.pug
+++ b/emails/welcome/html.pug
@@ -1,0 +1,22 @@
+//- Welcome email sent when person first signs in
+//- expects locals of {
+//-   to: person receiving email
+//-   from: person sending email - the op requestor
+//- }
+extends ../theme/layout.pug
+
+block tagline
+  span Welcome to Voluntarily
+
+block copy
+  p Kia ora #{to.nickname},
+  p Thanks for volunteering to help out. 
+  p We have setup your account and allocated your first set of goals. 
+  p Click the button below to open Voluntarily on your home page, complete your profile and start your training.
+  p Ngā mihi, Voluntarily.nz
+
+block copybutton
+  a(href=to.href class='mobile-button') Lets get things started
+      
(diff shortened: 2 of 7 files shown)


emails/theme/footer.pug
```diff
@@ -9,13 +9,11 @@ table(width="100%")
...
                           span This email was sent to #{to.email}. You're receiving this email because you expressed interest in Voluntarily Opportunities.
-                          span If you'd rather not see us in your inbox in the future, please set your
-                            a(href=to.href) profile to inactive.
-
-                    if (unsubscribeLink)
-                      p
-                        a(href=unsubscribeLink) Unsubscribe
+                      p If you'd rather not see us in your inbox in the future, 
+                        if (unsubscribeLink)
+                          a(href=unsubscribeLink) Unsubscribe
```
guttmann: Should we put the `p` inside the if statement too?

emails/welcome/html.pug
```diff
@@ -0,0 +1,22 @@
...
+  p Kia ora #{to.nickname},
+  p Thanks for volunteering to help out. 
+  p We have setup your account and allocated your first set of goals. 
+  p Click the button below to open Voluntarily on your home page, complete your profile and start your training.
+  p Ngā mihi, Voluntarily.nz
+
+block copybutton
+  a(href=to.href class='mobile-button') Lets get things started
+      
+block copyimage
```
guttmann: Is this just an empty block or is it doing something I'm not aware of?

server/api/person/person.email.js
```diff
@@ -12,52 +12,46 @@ module.exports.emailPerson = async (template, to, props, renderOnly = false) =>
   if (to._id && !to.sendEmailNotifications) {
     return null
   }
-  try {
```
guttmann: What impact does removing this try/catch block have?

I see your commit mentions the exception will run up to the request handler - I'm not 100% sure what this will mean for the user but I assume whatever request is in progress will essentially fail. That seems like a large change from the email not sending and the error being logged.

emails/welcome/html.pug
```diff
@@ -0,0 +1,22 @@
...
+
+block copy
+  p Kia ora #{to.nickname},
+  p Thanks for volunteering to help out. 
+  p We have setup your account and allocated your first set of goals. 
+  p Click the button below to open Voluntarily on your home page, complete your profile and start your training.
+  p Ngā mihi, Voluntarily.nz
+
+block copybutton
+  a(href=to.href class='mobile-button') Lets get things started
```
guttmann: What is the value of `to.href` here?

Could we use the globally available (within email templates) `appUrl` assuming this link just sends people to the homepage?

guttmann: You can probably ignore this I just found where it is being set further down

server/api/person/person.subscribe.js
```diff
@@ -0,0 +1,27 @@
...
+ * @param {object} to person email is for. (requestor or volunteer) with email populated.
+ * @param {object} props extra properties such as attachment
+ */
+const welcomeFrom = {
+  nickname: 'Welcome',
+  name: 'The team at Voluntarily',
+  email: 'welcome@voluntarily.nz'
+}
+
+module.exports = (server) => {
```
guttmann: I see `server` isn't used here - I'm assuming it's here because it's essentially an interface and `server` may be used in other `*.subscribe.js` functions?

emails/theme/footer.pug
avowkind: no its ok

emails/welcome/html.pug
avowkind: it overrides the default image and shows no image - will be given an image during content prettying.

server/api/person/person.subscribe.js
avowkind: thats correct

server/api/person/person.email.js
avowkind: I think a failure here implies serious problem with the email subsystem we should be returning 500. We need to make the system intolerant for a while.

PR #679, opened 2020-02-26
VP-1074 send emails when following or joining an organisation

avowkind:

## Proposed Changes? 🤔
1. When a new membership record is created, or updated publish a TOPIC_MEMBER__UPDATE message containing the new record - populated with person and org. - Note there are 3 places where this happens - create, update controllers and addMember in member.lib
2. Subscribe to the TOPIC_MEMBER__UPDATE message in person.subscribe and send an email to the person involved telling them about their new status.  ensure to and org links are canonical. 
3. add email templates of the form member_follower, member_member, member_orgadmin etc. for each of the state changes that generate emails. 

Note - the templates don't have final language set.

## Screenshots 📷
 
image


server/api/person/person.subscribe.js
```diff
@@ -24,4 +25,25 @@ module.exports = (server) => {
     })
     PubSub.publish(TOPIC_PERSON__EMAIL_SENT, info)
   })
+
+  PubSub.subscribe(TOPIC_MEMBER__UPDATE, async (msg, member) => {
```
guttmann: Should member specific subscriptions live in `server/api/member/member.subscribe.js`?

avowkind: I thought about that.  we put the subs where they are relevant to be handled rather than started. So while the publish might be in member or interest. the subs for adding goal cards are in personalGoals, and the subs for sending emails to people are in person.  If we added a transaction log that would have its own module and collect any topics it needed to log.

PR #681, opened 2020-02-27
VP-1071 send emails on interest status changes

avowkind:

## Proposed Changes? 🤔
When a person indicates interest in a volunteering opportunity a new interest record is created and this tracks changes as they are invited, commit, withdraw etc.  Once the event is completed the record becomes an archivedInterest record and the state may become attended or notattended.
On each state change the controller publishes a TOPIC_INTEREST__UPDATED message.  This is picked up by the person.subscribe.js module which generates emails to the volunteer, the op requestor or both depending on the state change.  emails are provided with the details of the opportunity, requestor, volunteer and interest status. 

New templates added for each state. templates are of the form interest_vp_state or interest_op_state. 

## More information
Emails on interest state changes were already present but sent directly from the interest controller.  this now uses the pub/sub method to dispatch the change and handle it later which will speed up the request response.


server/api/person/__tests__/person.email.spec.js
```diff
@@ -60,14 +59,14 @@ test('Send acknowledgeInterest email to person', async t => {
   t.regex(sentMail[0].subject, /Confirming your interest/)
 })
 
-test('render acknowledgeInterest email to person', async t => {
+test.only('render acknowledgeInterest email to person', async t => {
```
guttmann: Should this `.only` be here?

server/api/person/person.email.js
```diff
@@ -21,8 +22,8 @@ module.exports.emailPerson = async (template, to, props, renderOnly = false) =>
     send: true,
     subjectPrefix: config.env === 'production' ? false : `[${config.env.toUpperCase()}] `,
 
-    // Comment the line below to see preview email
-    preview: false,
+    // uncomment the line below to see preview email
+    // preview: true,
```
guttmann: Is leaving `preview` commented out equivalent to having `preview: false`?

server/api/person/__tests__/person.email.spec.js
avowkind: good catch.

server/api/person/person.email.js
avowkind: it is

PR #684, opened 2020-02-27
Vp467 improve org page

waltzaround:

## Proposed Changes? 🤔
1. small layout and copy fixes for orgdetail

## Screenshots 📷
 

image


components/VTheme/Profile.js
```diff
@@ -1,31 +1,56 @@
 import styled from 'styled-components'
 
 export const ProfileBanner = styled.header`
```
avowkind: I think we should rename profile banner to VBanner as we are using it consistently across all pages that have banners.

components/VTheme/Profile.js
```diff
@@ -1,31 +1,56 @@
...
+
 @media screen and (min-width: 767px) {
-  padding-left: 3rem;
+
   text-align: left;
 }
 
 `
+
+export const ProfileBannerImageContainer = styled.div`
```
avowkind: suggest rename to VBannerImage and make it a styled.img. Then you don't need to wrap the image in the styling but can just have <VBannerImage src=etc   cleaner code. 
This makes it clearer that we mean the top left image on the banner.

components/VTheme/Profile.js
```diff
@@ -1,31 +1,56 @@
...
-  grid-template-columns: 25% 75%;
+  grid-template-columns: 15rem 1fr;
+  gap: 5rem;
+ 
 }
 
 img {
-  width: 16rem;
-  max-height: 12rem;
+  width: 20rem;
```
avowkind: If we expect more than one image inside a banner this will affect both of them. Using the VBannerImage described below will distinguish it from any others.  This section may be unnecessarily duplicating styles.

components/VTheme/Profile.js
waltzaround: They do differ though - but will refactor for upcoming profile changes as well

components/VTheme/Profile.js
waltzaround: might style img inside the VBanner instead

components/VTheme/Profile.js
waltzaround: There aren't any cases where there's more than one image though

either way, this has been refactored out

components/VTheme/Profile.js
waltzaround: Sussed

waltzaround:
Re: removing page-specific styling thats slowly happening - this is reusing the ActivityContainer grid

PR #685, opened 2020-02-27
VP-1350 Clear idToken cookie on JWT expiry

guttmann:

## Proposed Changes? 🤔

1. Clear the idToken cookie if the JWT is found to be expired

## Additional Info.🧐

I've been finding that when I start work in the mornings on Voluntarily my JWT is always expired and I have been having to jump through hoops to sign back in (usually either clicking the "Sign out" button and then signing in again, or clearing my cookies before attempting to sign in). If I don't do this and just click "Sign in" straight away the first page I viewed would appear to be signed in successfully but as soon as I navigate to a different page I was ending up back in the signed out state.

This change removes that frustration - the first request after the JWT expires will now clear the idToken cookie and clicking on "Sign in" will sign you in as expected.

### DIFF
--- a/server/middleware/session/setSession.js
+++ b/server/middleware/session/setSession.js
@@ -74,6 +74,11 @@ const setSession = async (req, res, next) => {
     user = await jwtVerify(idToken)
   } catch (e) {
     // console.error('Jwt Verify failed', e)
+
+    if (e.name === 'TokenExpiredError') {
+      res.clearCookie('idToken')
+    }
+
     user = false
   }
   if (!user) {
--- a/server/middleware/session/__tests__/setSession.fixture.js
+++ b/server/middleware/session/__tests__/setSession.fixture.js
@@ -104,6 +104,25 @@ const jwtDataCharles = {
 }
 jwtDataCharles.idToken = jwt.sign(jwtDataCharles.idTokenPayload, 'secret')
 
+const jwtDataExpired = {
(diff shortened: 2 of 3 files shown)


guttmann:
I'm a bit unsure about the `accessToken` cookie too - @avowkind do you think that cookie should be cleared as well?

I assume that it's primarily used by Auth0 so if they decide it's expired they'll prompt the user to login again so maybe we don't need to worry about it.

avowkind:
> I'm a bit unsure about the `accessToken` cookie too - @avowkind do you think that cookie should be cleared as well?
TBH I don't know what the accessToken is there for. I don't think we ever test or use it. 
However if you are calling unsetToken to clear the cookies then we can group together any required changes.

server/middleware/session/setSession.js
```diff
@@ -74,6 +74,11 @@ const setSession = async (req, res, next) => {
     user = await jwtVerify(idToken)
   } catch (e) {
     // console.error('Jwt Verify failed', e)
+
+    if (e.name === 'TokenExpiredError') {
+      res.clearCookie('idToken')
```
avowkind: it might be safer to use unsetToken from auth.js.

guttmann: I was under the impression that unsetToken was only for use in browser (because it uses `js-cookie`) where as this code is always running on the server side

PR #686, opened 2020-02-27
Vp 1349 api call to request sending email message to a gro

avowkind:

## Proposed Changes? 🤔
This change adds a new messages array to the interest record.  this will replace the single comment string which is deprecated (but not removed for back-compatibility). 
A message contains body text, timestamp and author which allows messages to be left by both the op.requestor and interested volunteer.   Existing ability rules for interest records should allow both parties to perform updates.

While the ticket calls for a new API to enable emails to the interested group this method allows both on screen messages and emails to be associated with all state changes and updates and does not require a new api endpoint.

### Additional Information
Current behaviour. when a new interest record is created the volunteer can enter a comment. this is stored in the record and shown to the organiser when they review the interested people.

New behaviour.  comment field is kept but deprecated. to be replaced with messages array.
Each interest record holds a messages array.  Each message contains a body text, author and timestamp. 
When a new interest record is created an initial message from the volunteer is added. When there is a state change e.g - invited, committed, etc. the person initiating the state change can add a message. which is added to the array.  
At other times the two people may add another message. messages will be listed as a chat sequence on the page.

This PR only implements the database and API side supporting interested messages. No change to client side components.


server/api/interest/__tests__/interest.spec.js
```diff
@@ -205,14 +226,38 @@ test.serial('Should update the interest state from interested to invited', async
...
   t.is(res.status, 200)
-  t.is(res.body.status, 'invited')
+  const updateInterest = res.body
+  t.is(updateInterest.status, InterestStatus.INTERESTED)
+  t.is(updateInterest.messages.length, 2)
+  t.is(updateInterest.messages[1].name, from.nickname)
+})
+
+test.serial('Should update the interest state from interested to invited', async t => {
+  // interests 2 has a single date - so should trigger a calendar event to be attached
```
guttmann: This comment appears to have been copy and pasted and is not relevant here

server/api/interest-archive/__tests__/interestArchive.spec.js
```diff
@@ -21,14 +21,18 @@ test.after.always(async (t) => {
...
-  t.context.people = await Person.create(people).catch((err) => `Unable to create people: ${err}`)
+  t.context.people = await Person.create(people)
   archivedOps.map((op, index) => { op.requestor = t.context.people[0]._id })
-  t.context.opportunities = await ArchivedOpportunity.create(archivedOps).catch((err) => console.error('Unable to create opportunities', err))
+  t.context.opportunities = await ArchivedOpportunity.create(archivedOps)
   interests.map((interest, index) => {
     interest.opportunity = t.context.opportunities[index]._id
     interest.person = t.context.people[index]._id
+    interest.messages = [{ // this works whether its an object or array.
+      body: `${t.context.people[index]._id.name} is interested.`,
```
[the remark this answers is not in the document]
avowkind: you're right - I'll correct it, its a placeholder string so wouldn't affect the test.

server/api/interest/__tests__/interest.spec.js
avowkind: yes several times over - will remove

PR #693, opened 2020-03-02
Vp 1362 update registerinterestcomponent to use messages

avowkind:

## Proposed Changes? 🤔
Adds the ability to attach messages from the vp and op when interest record states change. 
When a Get Involved button is pressed or when a person accepts or rejects an invitation or commitment a popup message editor appears where they can leave a note for the organiser. This box also collects their acceptance of terms and conditions when first used. 

We track whether the person is moving to accept the positive button, rejecting via the negative button or simply sending a message. 

After the record has been updated we send an email to the organiser informing them of the new status and including the vp's message. 

We now have lots more email templates as different messages are sent depending on whether the op or vp is either accepting or rejecting the new status. 

The Current interested status is shown in an 'Action bar' across the top of the page. this contains the current call to action along with the relevant accept, reject or message buttons.

## Screenshots 📷
 Screen Shot 2020-03-02 at 10 13 11

Screen Shot 2020-03-02 at 10 14 23

Screen Shot 2020-03-02 at 10 09 18

Screen Shot 2020-03-02 at 10 08 59

Screen Shot 2020-03-02 at 10 09 09


avowkind:
Note to reviewers:  the body text in the email templates is mainly illustrative and is not the final copy.  I've tried to word it so it's clear which type of transition the email represents but it will definitely need to change in a later ticket.

waltzaround:
Going to merge @avowkind ?

PR #696, opened 2020-03-03
VP-1222 Once an idToken expires and fails validation we have to redirect to sign in again

wseagar:

## Proposed Changes? 🤔
1. In all cases - server side or client side, if the authentication token is expired the user will be asked to reauthenticate and then be returned to the page they were on.
~~2. Edge case - if the token expires while the browser is making an AJAX request (unlikely) they will be redirected to the home route.~~
2. If the token expires while the browser is making an AJAX request to a /api/ route then the server will respond with a 401 unauthorized
3. The sign-thru react component always clears cookies, this is so that if the user doesn't complete the authentication process. They will always come back to vly as a unauthorised user (not redirected again to reauth).


server/middleware/session/setSession.js
```diff
@@ -75,8 +77,26 @@ const setSession = async (req, res, next) => {
...
 
-    if (e.name === 'TokenExpiredError') {
-      res.clearCookie('idToken')
+    if (e instanceof TokenExpiredError) {
+      // Don't redirect user to an /api/ path after signing through,
+      // the user would be treated with a nice wall of json.
+      // This should only happen if the access token expires after page load but before ajax is exectued on a page.
+      const redirectUrl = req.url.startsWith('/api/') ? '/home/' : req.url
+
+      const qs = queryString.stringify({ redirect: redirectUrl })
```
wseagar: Unsure if queryString library is the right choice. I saw it being used elsewhere.

encodeURIComponent would work also.

lib/auth/__tests__/auth.spec.js
```diff
@@ -96,3 +97,47 @@ test.serial('getSession from cookies', async t => {
...
+  global.window = {
+    location: {
+      protocol: 'https:',
+      host: 'beta.voluntarily.nz',
+      pathName: '/home',
+      search: '?tab=history',
+      hash: '',
+      replace: fakeLocationReplace
+    }
+  }
```
wseagar: This mock window implementation feels kind of hacky. I'm not sure if theres a better way.

avowkind: we can mock the router class but for a simple test this is clear enough.

lib/auth/auth.js
```diff
@@ -44,20 +46,44 @@ export const getSessionFromCookies = async () => {
...
+  // Check if token is not expired on client side.
+  // This may happen if the users token expires while the are navigating between pages that
+  // are not server side rendered.
+  // NOTE: This does not check if the token has been signed by the correct signing key,
+  // we do not need to verify the JWT on the client side as if an attacker were forging a signed JWT
+  // the server will reject all API and page load requests.
+  const jwtValid = isJwtExpValid(user.exp)
+  if (!jwtValid) {
+    const redirectUrl = encodeURIComponent(getPathAndHash())
+    window.location.replace(`${getBaseUrl()}/auth/sign-thru?redirect=${redirectUrl}`)
```
avowkind: This is simpler using the Router 
import Router from 'next/router'
    const redirectUrl = encodeURIComponent(getPathAndHash())
    const signThruUrl = `/auth/sign-thru?redirect=${redirectUrl}`
    Router.push(signThruUrl)

lib/auth/auth.js
```diff
@@ -44,20 +46,44 @@ export const getSessionFromCookies = async () => {
...
   }
   return session
 }
 
+/**
+ * Takes a exp time from a JWT and returns if it has not expired.
+ * @param {number} exp Expiration time claim from a JWT
+ * @returns {boolean} returns true if the token is valid
+ */
+export const isJwtExpValid = (exp) => {
```
avowkind: Cleaner to use moment both times. 

moment() > moment(exp)

lib/urlUtil.js
```diff
@@ -20,6 +20,19 @@ const defaultToHttpScheme = (url) => {
   return url
 }
 
+// https://nodejs.org/api/url.html#url_url_strings_and_url_objects
+// contains a good graphical representation of the URL specification
+const getBaseUrl = () => `${window.location.protocol}//${window.location.host}`
```
avowkind: these extra Utils not required as functionality is in the Next Router.
import { useRouter } from 'next/router'
const router = useRouter()
{router.asPath}. instead of getPathAndHash

server/middleware/session/setSession.js
avowkind: yes please use encodeURIComponent

server/middleware/session/setSession.js
```diff
@@ -75,8 +77,26 @@ const setSession = async (req, res, next) => {
   } catch (e) {
     // console.error('Jwt Verify failed', e)
 
-    if (e.name === 'TokenExpiredError') {
-      res.clearCookie('idToken')
+    if (e instanceof TokenExpiredError) {
+      // Don't redirect user to an /api/ path after signing through,
+      // the user would be treated with a nice wall of json.
+      // This should only happen if the access token expires after page load but before ajax is exectued on a page.
+      const redirectUrl = req.url.startsWith('/api/') ? '/home/' : req.url
```
guttmann: Won't this result in API requests made from the client (after the token has expired) resulting in a 200 response that contains the home page HTML?

Would it be better in the event we find the token has expired and the current request is to the API to give a response that the client would be able to identify as such and handle appropriately (e.g. send a 401 status code)?

wseagar: My bad, I think you are partially correct. By default I don't think the redirect will be followed with fetch. It depends on what HTTP abstraction your using in the browser.  xmlhttprequest will follow redirects but fetch doesn't. 

All that would be needed to make this work would be to add "redirect:follow" to the fetch object on the front end https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#Supplying_request_options

A 401 would be more appropriate. However it will need to be handled client side. In my manual testing it appeared the requests just fail. (I haven't looked at the code yet)

lib/urlUtil.js
wseagar: useRouter hook only works inside the body of a functional react component.

Just tried it and get the following error:

```
  Error {
    message: `Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:␊
    1. You might have mismatching versions of React and the renderer (such as React DOM)␊
    2. You might be breaking the Rules of Hooks␊
    3. You might have more than one copy of React in the same app␊
    See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.`,
  }
```

server/middleware/session/setSession.js
wseagar: Hi James, I have changed the code to return a 401 status code. I'm not sure what the client side behavior will be.

lib/urlUtil.js
avowkind: ah ok.  I'm sure there's a server side version but I'll pull this in for now.

PR #701, opened 2020-03-05
VP-1323 Volunteer interest update

guttmann:

## Proposed Changes? 🤔

Implement permissions around interest status transitions for the interest API. Primarily this applies to people with the "volunteer" role who should only be able to update status from:
- "invited" -> "committed"
- "committed" -> "interested"

### DIFF
--- a/server/api/interest/interest.controller.js
+++ b/server/api/interest/interest.controller.js
@@ -3,6 +3,8 @@ const { Action } = require('../../services/abilities/ability.constants')
 const { getInterestDetail, getMyOpInterestDetail } = require('./interest.lib')
 const { TOPIC_INTEREST__UPDATE, TOPIC_INTEREST__MESSAGE, TOPIC_INTEREST__DELETE } = require('../../services/pubsub/topic.constants')
 const PubSub = require('pubsub-js')
+const { Role } = require('../../services/authorize/role')
+const { InterestStatus } = require('./interest.constants')
 
 /**
   api/interests -> list all interests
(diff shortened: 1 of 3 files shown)


server/api/interest/interest.controller.js
```diff
@@ -102,25 +104,51 @@ const createInterest = async (req, res) => {
...
+      }
+    }
+
+    if (interestUpdateData.status) {
+      existingInterest.status = interestUpdateData.status
+    }
+
+    if (interestUpdateData.messages) {
+      interestUpdateData.messages = Array.isArray(interestUpdateData.messages)
+        ? interestUpdateData.messages : [interestUpdateData.messages]
```
avowkind: messages should always be an array. even if usually it contains only one message. but this is a safe action.

guttmann: > messages should always be an array. even if usually it contains only one message. but this is a safe action.

I found code & comments that contradict that -

Some of the tests in `server/api/interest/__tests__/interest.ability.spec.js` send messages as an array, e.g.

https://github.com/voluntarily/vly2/blob/9a702d5e80b02ae96f1dc140a093f23718b61b63/server/api/interest/__tests__/interest.ability.spec.js#L206-L209

And one of the tests in `server/api/interest/__tests__/interest.spec.js` has a comment mentioning it could be either:

https://github.com/voluntarily/vly2/blob/9a702d5e80b02ae96f1dc140a093f23718b61b63/server/api/interest/__tests__/interest.spec.js#L216-L219

If the intent is for it to always be an array maybe we should update these areas to match that?

PR #706, opened 2020-03-06
VP-1347 Op Manager can message volunteers

avowkind:

## Proposed Changes? 🤔
1. Teachers (op requestors) can message volunteers from the Management tab of the opportunity page. 
2. Archived Opportunity page now much more like the opportunity page.

Lots of smaller changes
interestsArchived renamed to interestArchives to match the underlying collection
interestArchived api merged into interest api folder. only the ability and router files are different. 

These components recoded as functions simplified and support for messages added

- InterestSection
- InterestTable
- InterestArchivedSection
- InterestArchivedTable

Action Buttons made more generic handling Accept, Reject and Message actions with different labels depending on context.

Opportunity Requestor Manage tab now allows the teacher to expand a view of the person and see an exchange of messages from both sides

Shared interest fixture for Ability tests.
Fixed 
- sending messages to the wrong person
- people with mixed roles e.g. vp+op+ap can now do stuff like volunteer.

Still to do (other tickets)
- expanded row in InterestTable uses PersonCard which needs improving.
- Volunteer does not have a place to view the message history.
- Still some strings to be put in translation - will do separate sweep
- History page has InterestArchivedTable visible to all people - should be op and op admin only.
- But trying to use the buttons is blocked by api.
- Need an Icon to show attended was marked on history ops.
- change from use type accept to select the emails - use the old and new state transition as this will allow more specific emails.
- not convinced the termsAccepted is being saved - the message appears most of the time.
- see TypeError: op.tags is undefined on some new active activities. SSR only
- sender name not shown in message stack.

### Original
image

### Updated
image

image


waltzaround:
Reviewing in a mo

PR #707, opened 2020-03-07
VP-1379 create generic closing state

waltzaround:
- [] Not decreased the overall test coverage?

## Proposed Changes? 🤔
1. created dumb activity creation close component
2. created dumb org school creation close component 

## Screenshots 📷
 
image
image


avowkind:
Walt these are basically the same options as the goal cards.  Do you think that the panels can be generated as a different view component on the goal cards?  These are not the sort of thing we want hard coded?

waltzaround:
@avowkind Agree but how do we want to store the markup?

avowkind:
The card is in the database we can store html there, we can add more fields if we want a long or short form of the card. 

I’m concerned that clicking these mini-cards bypasses the features of the goal cards - the in progress, the instructions, the completion tests etc. 

Have you considered reworking the personal goals section of the home page to look like this?

waltzaround:
I dont think these need to be as complex as goalcards imho - not to mention most of the actions on the minicards are very small (ie: share the event) and do not need to be tracked

PR #712, opened 2020-03-09
Vp 580 capture my org domain name so that i can identify members by email

Eteroa123:

## Proposed Changes? 🤔
1. Created a organisation validation and added the regex rules
2. Added a Domain name for a organisation form
3. Made sure it was a valid domain name
4. Even though its not meant to display manage to save it on the form 

## Screenshots 📷
 

Screen Shot 2020-03-09 at 11 52 44 PM

Screen Shot 2020-03-09 at 11 54 35 PM


Eteroa123:
OHHHHH MAAAAA GAWSH !!!!!!!!!!! i'll fix that in the morning

components/Org/OrgDetailForm.js
```diff
@@ -40,6 +41,10 @@ class OrgDetailForm extends Component {
   setAddress = value => {
     this.props.form.setFieldsValue({ address: value })
   }
+
+  /* setDomainName = value => {
```
avowkind: remove commented out code

components/Org/OrgDetailForm.js
```diff
@@ -321,6 +340,18 @@ class OrgDetailForm extends Component {
                   <RichTextEditor />
                 )}
               </Form.Item>
+              <ShortInputContainer>
+                <Form.Item label={orgDomain}>
+                  {getFieldDecorator('domainName', {
+                    rules: [
+                      {
+                        pattern: domainRegex,
+                        message: 'Enter vaild domain'
```
avowkind: spelling s/vaild/valid

server/api/person/person.validation.js
```diff
@@ -1,15 +1,24 @@
 const websiteRegex = /^(?:http(s)?:\/\/)?[\w.-]+(?:\.[\w.-]+)+[\w\-._~:/?#[\]@!$&'()*+,;=.]+$/
+const domainRegex = /(?:[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?\.)+[a-z0-9][a-z0-9-]{0,61}[a-z0-9]/
```
avowkind: person.validation and organisation.validation appear to be the same file.  Can we have just one called fieldValidation.js?

Eteroa123: @avowkind oh where would "fieldValidation.js" be stored. 
I forgot to remove the domain regex on person.validation.js
But would it be better to keep the person.validation and organsation.validation separate ????

Eteroa123: Let me know what you think  @avowkind

avowkind: the rules should be the same for every form that uses them so put them all together in one validation file.

components/Org/OrgTabs.js
```diff
@@ -89,7 +89,7 @@ OrgTabs.propTypes = {
     imgUrl: PropTypes.string,
     website: PropTypes.string,
     contactEmail: PropTypes.string,
-    facebook: PropTypes.string,
+    domainName: PropTypes.string,
```
bernesemountaindog: Should `facebook` have been replaced here?

server/api/organisation/organisation.controller.js
```diff
@@ -45,8 +45,13 @@ const putOrganisation = async (req, res) => {
 
   await Organisation.findByIdAndUpdate(req.params._id, { $set: req.body })
   res.json(await Organisation.findById(req.params._id).exec())
+  // Domain string validation
```
bernesemountaindog: Upon reading this I realise the way I originally did this (in `person.controller.js`) could be done better, by simply calling the validation function itself as opposed to using the regex.

Make use of the validation method:
```const orgValidation = require('./organisation.validation')
...
if (Object.keys(req.body).includes('domainName') && !orgValidation.domainName(req.body.domainName)) {
  return res.status(400).send('The \'domainName\' field does not match the validation rule')
}
```

server/api/organisation/organisation.controller.js
```diff
@@ -45,8 +45,13 @@ const putOrganisation = async (req, res) => {
...
+  if (Object.keys(Organisation).includes('domainName')) {
```
bernesemountaindog: Is `Object.keys(Organisation)` meant to be `Object.keys(req.body)`?

server/api/organisation/organisation.validation.js
```diff
@@ -0,0 +1,24 @@
+const websiteRegex = /^(?:http(s)?:\/\/)?[\w.-]+(?:\.[\w.-]+)+[\w\-._~:/?#[\]@!$&'()*+,;=.]+$/
```
bernesemountaindog: As the `regex` for validating an organisation's website is now here, should `OrgDetailForm` be updated to using this regex to avoid duplication?

bernesemountaindog: Could we also extract this `regex` and share a common one between this new file and `person.validation.js`?

Eteroa123: Oh sweet that makes sense to take it out !!!

server/api/person/person.validation.js
avowkind: > @avowkind oh where would "fieldValidation.js" be stored.
/lib is for stuff everyone uses.

components/Org/OrgDetailForm.js
```diff
@@ -24,7 +25,7 @@ function hasErrors (fieldsError) {
 class OrgDetailForm extends Component {
   constructor (props) {
     super(props)
-    this.setImgUrl = this.setImgUrl.bind(this)
+    // this.setImgUrl = this.setImgUrl.bind(this)
```
avowkind: Why is this commented out?

server/api/organisation/organisation.controller.js
Eteroa123: @bernesemountaindog  So i added the new line of code that you gave to the org controller. 
I'm going to make one validation file so that it can cover the person and organisations

components/Org/OrgDetailForm.js
Eteroa123: I uncommented this.setImgURL

server/api/organisation/organisation.validation.js
Eteroa123: I removed the website regex in the organisation validation.js

server/api/organisation/organisation.controller.js
Eteroa123: Made the change to res.body

server/api/organisation/organisation.controller.js
```diff
@@ -46,7 +46,7 @@ const putOrganisation = async (req, res) => {
   await Organisation.findByIdAndUpdate(req.params._id, { $set: req.body })
   res.json(await Organisation.findById(req.params._id).exec())
   // Domain string validation
-  if (Object.keys(Organisation).includes('domainName') && !orgValidation.domainName(req.body.domainName)) {
+  if (Object.keys(res.body).includes('domainName') && !orgValidation.domainName(req.body.domainName)) {
```
avowkind: should be **req.body** not res. body. 

the idea is that you are about to call orgValidation.domainName on the request body domainName.  But you don't want to do this if it is not in the request.  So you might want to check req.body.domainName is not Null.  but if the field is not present this would throw an exception.  So instead we check whether it is in the object keys. Thats the list of properties that the object actually has.

Eteroa123:
I'll work on now creating a individual validation file !!!!!!

PR #717, opened 2020-03-11
VP-597 Organisation opportunity history tab

guttmann:

## Proposed Changes? 🤔

1. Add "History" tab to organisation detail page for opportunity provider organisations
2. Add "scenario" db script that loads data into dev environment to make development/testing easier
3. Add new `mountWithMockIntl` function that lets us provide custom intl messages during unit tests
4. Add new CI task to run `build:lang` and fail if differences are found between the committed `lang/en.json` file and the freshly generated one.

### Scenario script

I didn't want to manually create a new organisation and then create a bunch of opportunities and transition those to archived opportunities - so I wrote a script that inserts the necessary data straight into the database. It requires an email address be provided for an existing person record (to give that person access to the organisation that gets created).

Usage: `node x/db/scenarios/opportunity-history.js test@example.com`

This will create a new organisation, a member record for the given person/email, and 10 archived opportunities.

### mountWithMockIntl

I've added this new function so that where we are referencing content in tests we don't have to hardcode the default value used in the formatted message but can rather mock that value for the specific test being run and use that value to test against.

### New CI task to check lang file state

I've added a new CI task that regenerates the `lang/en.json` file, the task follows these steps:

1. Make a copy of `lang/en.json` at `lang/en.orig.json`
2. Run `npm run build:lang`
3. Compare the newly generated `lang/en.json` file against `lang/en.orig.json`, if there are differences found the `diff` command has an exit code of 1 which will cause the build to fail. If there are no differences then the `diff` command exits with a 0 and the build passes.

In the event this task fails during the build, run `npm run build:lang` and commit the updated `lang/en.json` file.

### Original

Imagine the organisation detail page without a tab that says "History"

### Updated

image


components/Org/OrgTabs.js
```diff
@@ -1,73 +1,58 @@
...
-    defaultMessage='Edit'
-    description='Tab label for org Editor panel on organisation page'
-  />
+const { TabPane } = Tabs
 
 // Warning do not try to group tabs under an isFlag TabPanes must be direct Children of Tabs.
-export const OrgTabs = ({ org, onChange, canManage, defaultTab, isAuthenticated }) => (
+export const OrgTabs = ({
+  org,
+  archivedOpportunities,
```
avowkind: this is being passed down two levels of component when it is already in Redux. 
its not necessary to do this. 
using either withArchivedOpportunities or useSelect in the panel itself you can get access to the data.

__tests__/org/orgdetailpage.spec.js
```diff
@@ -76,6 +76,21 @@ test.before('Setup fixtures', (t) => {
...
+      syncing: false,
+      loading: false,
+      data: [{
+        name: 'test',
+        subtitle: 'test',
+        imgUrl: 'test.jpg',
+        date: ['2017-09-03', '2013-07-26'],
+        location: 'Testington',
+        duration: 'long time',
+        _id: 'testidtestid'
```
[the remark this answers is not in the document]
guttmann: From the perspective of rendering in an `<OpList />` it doesn't appear to be required (I based the fields here off the PropTypes definition

pages/org/orgdetailpage.js
```diff
@@ -171,6 +181,7 @@ OrgDetailPage.getInitialProps = async ({ store, query }) => {
       orgid: null
     }
   } else if (query && query.id) {
+    await store.dispatch(reduxApi.actions.archivedOpportunities.get({ q: JSON.stringify({ offerOrg: query.id }) }))
```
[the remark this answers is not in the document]
guttmann: Have addressed this in my latest changes by moving the redux stuff into the `<OrgHistoryPanel />`

components/Org/OrgTabs.js
guttmann: Also resolved in my latest changes by moving redux stuff into the `<OrgHistoryPanel />`

components/Org/OrgHistoryPanel.js
```diff
@@ -0,0 +1,58 @@
...
+        reduxApi.actions.archivedOpportunities.get({ q: JSON.stringify({ offerOrg: organisationId }) })
+      )
+    }
+  }, [organisationId])
+
+  let content = ''
+
+  if (loadingState.loading || (!loadingState.sync && !loadingState.error)) {
+    content = <ReduxLoading />
+  } else if (loadingState.error) {
```
avowkind: Why bypass and then reproduce the behaviour of <Loading> ?

components/Org/OrgTabs.js
```diff
@@ -1,73 +1,51 @@
...
+export const OrgTabs = ({
+  org,
+  onChange,
+  canManage,
+  defaultTab,
+  isAuthenticated
+}) => (
   <VTabs defaultActiveKey={defaultTab} onChange={onChange}>
-    <TabPane tab={orgTab} key='about'>
+    <TabPane tab={orgTab} key='about' orgTab='about'>
```
avowkind: what is the purpose of the extra orgTab property?

components/Org/OrgTabs.messages.js
```diff
@@ -0,0 +1,53 @@
...
+  />
+
+const orgInstructionTab =
+  <FormattedMessage
+    id='orgInstructions'
+    defaultMessage='Getting Started'
+    description='show opportunities list on volunteer home page'
+  />
+
+const orgOffersTab =
```
avowkind: no need to do this now - but see OpStatus for a more compact way to define a set of translated strings.

components/Org/OrgTabs.js
guttmann: I needed to be able to get a specific `<TabPane />` in unit tests - I first tried using `wrapper.find('TabPane[key="history"]')` but it turns out you can't use `key` in attribute selectors. This is covered briefly in the Enzyme documentation, see the "React Key and Ref Props" section on this page: https://enzymejs.github.io/enzyme/docs/api/selector.html specifically:

> While in most cases, any React prop can be used, there are exceptions. The key and ref props will never work; React uses these props internally.

So I added the `orgTab` attribute so I could get the history `<TabPane />`. I added the `orgTab` attribute to the other tab panes as well for consistencies sake.

avowkind:
I wondered if it was for test.  Definitely safer than trying to use numbers,  
Thats all good.

components/Org/OrgHistoryPanel.js
guttmann: I didn't really think about it at the time but trying to use the `ReduxLoading` component to handle the error now the main problem I'm seeing is that when the `OrgHistoryPanel` component first renders the initial state of the Redux store for archivedOpportunities is:

    { sync: false, syncing: false, loading: false, error: undefined, data: [] }

So the `ReduxLoading` component displays the "not synced" message for a split second before the Redux store updates the `loading` flag to `true`.

In the app this appears as a flash of a warning message, followed by the loading animation, and finally the records show up, e.g.

vly-org-history-warning-flash

I'd like to avoid that flash of a warning message - especially in this case where it's always going to happen due to the initial state of the redux store. I appreciate the message only shows in development environments but I don't think it should be shown at all when things are working as intended.

avowkind:
I found we needed the dev flash of the not loading state for a while when we were having sync problems.  
You can comment it out now so that a dev would have to turn it on for debugging.  

A.

PR #722, opened 2020-03-12
VP-1371 Volunteer can see messages about the Activity they are involved with.

avowkind:

## Proposed Changes? 🤔
As a volunteer viewing an opportunity details page (an activity) which I've got involved with. 
under the banner I can see the latest message from either myself or the organiser. 
Clicking on this message pops up a list of all the messages created during the invitation and acceptance process. The popup self clears after 5 seconds. It is scrollable although there shouldn't be many messages. 

## Additional Info.🧐
Messages are only updated when the page is refreshed.  This is not a chat window so you won't see an update if the other person does something.  - theres a later story to refresh the page automatically.

### Original
image

### Updated
image
image


waltzaround:
Awaiting test

PR #733, opened 2020-03-13
VP-1395 Improve State tags on OpCard

waltzaround:

## Proposed Changes? 🤔
1. Improve tags for OpCards

## Screenshots 📷

image

### DIFF
--- a/components/Op/OpCard.js
+++ b/components/Op/OpCard.js
@@ -67,13 +67,23 @@ const OpCard = ({ op }) => {
               {draft}
               {op.name}
             </h1>
-            {orgName}
+
             <p> {startLocation}</p>
             <p> {startTime} </p>
             <p> {startDuration}</p>
             <DescriptionWrapper>
               {op.subtitle}
             </DescriptionWrapper>
+            {orgName &&
+              <>
+                <DescriptionWrapper>
+
+                  <i>
+              By {orgName}
+                  </i>
+                </DescriptionWrapper>
+              </>}
+
           </figcaption>
         </a>
       </Link>
--- a/components/Op/OpCard.js
+++ b/components/Op/OpCard.js
@@ -23,10 +23,22 @@ const ImageWrapper = styled.div`
 `
 
 const StyledIcon = styled(Icon)`
-  font-size: 2rem;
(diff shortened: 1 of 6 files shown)


components/Op/OpCard.js
```diff
@@ -40,10 +40,10 @@ const OpCard = ({ op }) => {
   const interestIcon = ((interest) => {
     if (!interest) { return '' }
     switch (interest.status) {
-      case 'interested': return <StyledIcon type='like' theme='twoTone' twoToneColor='#6549AA' />
-      case 'invited': return <StyledIcon type='message' theme='twoTone' twoToneColor='#fb0' />
-      case 'committed': return <StyledIcon type='check-circle' theme='twoTone' twoToneColor='#0f0' />
-      case 'declined': return <StyledIcon type='close-circle' theme='twoTone' twoToneColor='#f00' />
+      case 'interested': return <TagState style={{ color: '#222', backgroundColor: '#E1E1E1' }}><StyledIcon type='mail' />You offered to help</TagState>
+      case 'invited': return <TagState style={{ color: 'white', backgroundColor: '#653CAD' }}><StyledIcon type='calendar' />You are invited</TagState>
+      case 'committed': return <TagState style={{ color: 'black', backgroundColor: '#36F482' }}><StyledIcon type='check-circle' />Accepted</TagState>
```
avowkind: At some point we can move these colour definitions to the theme.

components/Op/OpCard.js
```diff
@@ -40,10 +40,10 @@ const OpCard = ({ op }) => {
...
+      case 'declined': return <TagState style={{ color: 'white', backgroundColor: '#F44336' }}><StyledIcon type='close-circle' />Cancelled</TagState>
```
avowkind: declined doesn't really mean that the op was cancelled. it may be going ahead but the person has been rejected.

waltzaround: Yeah, we should probably come up with some better language

components/Op/OpCard.js
waltzaround: Can do - just need to move the palette from storybook into css variables

PR #765, opened 2020-03-21
Minor updates to getting started guide (on Mac)

christiedavis:
Very very minor updates to the getting started doc, to make it easier for the next person.


waltzaround:
Thanks for the fix @christiedavis

PR #775, opened 2020-03-21
[WIP] VP-1451 Production Monitoring Setup

AlexanderNZ:
This PR covers VP-1451 and subtasks.
There are a number of subtasks that still need to be addressed here, and further technical work to make the feature robust.

I'm using this PR to make the work transparent and collect feedback as early as possible.
Here's a rough list of technical things we need to address before this can be merged:

- Correctly implement and test both server and client solutions for crash reporting and RUM
- Plan and implement APM if available
- Ensure that both server and client have appropriate environment config, so that we can differentiate errors / traffic from different environments
- Ensure that we update bug specific workflows to send reports from users to the monitoring provider (instead of or in addition to Jira Service Desk)
- Configure appropriate notifications e.g. slack, email, in dire situations cellphone
- Onboard a bunch of users so that these tools can actually be used by the folk who need them
- Probably more to come!

## Proposed Changes? 🤔
1. Implement Raygun on the Server
2. Implement Statuspage
3. Document remaining AWS based work as well as notification flows

## Additional Info.🧐

see:
- https://voluntarily.atlassian.net/browse/VP-1466
- https://voluntarily.atlassian.net/wiki/spaces/VP/pages/220168272/Understanding+Run+State+of+the+Software


pages/_app.js
```diff
@@ -8,6 +8,9 @@ import { ThemeProvider } from 'styled-components'
 import { getSession } from '../lib/auth/auth'
 import { Role } from '../server/services/authorize/role'
 import { IntlProvider } from 'react-intl'
+import * as Sentry from '@sentry/browser'
+
+Sentry.init({dsn: "https://e8e41423c54e44218441559971b0a9d4@sentry.io/5169743"});
```
avowkind: Is this value ok to be in a public repository?  if not we can inject using encrypted env var.

AlexanderNZ: oops! have removed the sentry kit in favor of raygun, but have made the same mistake with the raygun api key. I'm so used to working on private repos, need to slow down a bit!

Have replaced the value and rotated the api key.  We should inject these values :)

server/api/person/person.routes.js
```diff
@@ -46,4 +46,8 @@ module.exports = function (server) {
       ]
     })
   )
+
+  server.get('/api/test/raygun', function (req, res) {
+    throw new Error()
+  })
```
AlexanderNZ: Note that this is hear for testing, and will be removed once we're happy that everything works as it should.

Dockerfile
```diff
@@ -5,6 +5,9 @@ WORKDIR /usr/src/app
 ENV PORT 3122
 EXPOSE 3122
 
+ENV RAYGUN_EXPRESS_SERVER_API_KEY=${RAYGUN_EXPRESS_SERVER_API_KEY}
```
avowkind: Wonder if there should be ARG statements here too.

AlexanderNZ: resolved in 0128450 :)

AlexanderNZ:
As part of https://voluntarily.atlassian.net/browse/VP-1466 I've written https://voluntarily.atlassian.net/wiki/spaces/VP/pages/220168272/Understanding+Run+State+of+the+Software

Important note that there are two new tools here, raygun and statuspage.

@avowkind - there are some options here:
Both these products are on trials, raygun pending a conversation with JD / Raygun and statuspage pending a subscription decision.

I went with statuspage because it's owned by atlassian and integrates well. Also since you have credits for atlassian I figured the same could apply to status page. The Hobby plan is $30USD per month so shouldn't burn through your credits or anything.

I've also linked all of these tools into Slack as per the docs I've written above.

I've aimed to make this stuff useful without making a large amount of decisions for you as I haven't been here long and am not going to be the final consumer. Would recommend the core voluntarily team have a sit down and work through some examples of when things go wrong, what happens?

The tools here support that, but won't make any sense without that overarching discussion!

As it stands, this is ready for review and merge. Note that Raygun will work for another 14 days without out input, Statuspage won't be visible until the atlassian conversation is had. Consider avoiding merging/releasing this work until the statuspage has been activated :)

PR #800, opened 2020-03-23
Bugfix/vp 1552 org groups list doesnt fetch existing groups before editing

MitchWeber:

## Proposed Changes? 🤔
1. 
2. 
3. 

## Additional Info.🧐

Any additional info goes here

## Screenshots 📷
Screen Shot 2020-03-23 at 10 34 43 PM


pages/org/orgdetailpage.js
```diff
@@ -23,7 +24,8 @@ const blankOrg = {
   website: null,
   facebook: null,
   twitter: null,
-  category: ['vp']
+  category: ['vp'],
+  tags: { data: [] }
```
avowkind: this doesn't look quite right. a blankOrg shouldn't have any tags. 
groups: []  perhaps.
Although likely not used at all.

MitchWeber: You are right. I changed it to `groups: []`

PR #843, opened 2020-03-27
VP-574 add womens refuge button

avowkind:

## Proposed Changes? 🤔
REplaced the womens refuge button in the footer.
This needed major rework - we now don't rely on the source scripts which failed during server side rendering.  Instead we just show our own iframe with the remote content provided. Using react open and close to control visibility.

This is not Modal . it just expands into the footer. which means that you can scroll away from it if necessary.

The large close button is not quite aligned.

## Additional Info.🧐

Any additional info goes here

## Screenshots 📷
 
image

image


waltzaround:
Reviewing now

PR #876, opened 2020-04-03
WIP:VP-1533 update robots.txt for prod vs non-prod

rushminatorr:

## Proposed Changes? 🤔
1. have production robots.txt allow home path for search crawlers

## Additional Info.🧐

Any additional info goes here
This PR includes two options, need to decide on one before merging. Please do not merge PR.

### Updated
OPTION 1: Have robots.txt specifically for prod override the non prod already existing file
OPTION2 : use node server to create robots.txt based on environment


Dockerfile
```diff
@@ -13,6 +13,7 @@ ENV RAYGUN_REACT_CLIENT_API_KEY=${RAYGUN_REACT_CLIENT_API_KEY}
 FROM base as production_build
 ENV NODE_ENV=production
 COPY . .
+COPY robots.txt.prod public/robots.txt
```
avowkind, line 16: This will happen on all deployments - alpha, beta, live etc.  as they are all production environments. (as opposed to dev or test).  Dev is only localhost and test is only when running npm test.   So this option won't work

server/server.js
```diff
@@ -75,6 +75,19 @@ const appReady = app.prepare().then(() => {
     next()
   })
 
+  // robots.txt
+  if (process.env.NODE_ENV === 'production') {
```
avowkind, line 79: This is a good option. However, the test should be for process.env.APP_URL includes ('live')

server/server.js
```diff
@@ -75,6 +75,19 @@ const appReady = app.prepare().then(() => {
...
+    server.get('/robots.txt', function (req, res) {
+      res.type('text/plain')
+      res.send('User-agent: *\nDisallow: /\nAllow:/home')
```
avowkind, line 82: You will want to allow most non signed in pages

server/server.js
asomervell: Really? What if someone else deploys the platform to not live?

avowkind:
Hi Rashi - are you able to complete this today? if not I'll do it.

server/server.js
rushminatorr: deferring to use allow as its not a standard and only googlebot knows about it.

server/server.js
rushminatorr: if not live - then don't allow anything to be crawled. 'Disallow: /'

rushminatorr:
I have accidentally rebased master on this branch at some point and messes up the history so deleting this branch and new PR created. Apologies!

PR #879, opened 2020-04-03
VP-1721 kick off a commit to a branch

asomervell:
Create a draft PR to push commits to


.github/workflows/push-master.yml
```diff
@@ -1,67 +1,80 @@
...
-        ci_node_index: 0
-        ci_node_total: 1
-    - name: Full test suite
-      run: |
-        npm ci
-        npm test
-
   docker-build:
     name: Build and push docker container
     runs-on: ubuntu-latest
```
avowkind, line 14: probably not important but ubuntu is pretty big - for a docker build there is probably a minimal Linux like Alpine that will do the job.

asomervell: haha it automatically says I "dismissed" a review! I didn't, but we covered this on call.

asomervell:
Ok @avowkind good to go if you want to review / merge. This will run a merge step that will deploy.... allegedly :p

PR #949, opened 2020-04-21
As a Volunteer/Activity Provider/Organisation I want to be able to click on the " See all Activity" and taken to the Activities page.

Eteroa123:
- [ *] Run `npm test` and all the tests pass.
- [* ] Run `npm run lint` and there are no warnings.
- [* ] Not decreased the overall test coverage?
- [* ] Included the Jira ID and description in the PR title

## Proposed Changes? 🤔
1. Both "See all activities "  are not connected to the activity page 

2. Both goal cards are fixed in the landing page. 

This PR is linked with 
  **VP-1760**


components/LandingPageComponents/GiveSupportSection.js
```diff
@@ -84,7 +84,7 @@ const GiveSupportSection = () => (
     <HeroSectionHeader>
       <h2>How you can support others</h2>
       <HeroSectionButtonContainer>
-        <Button type='secondary' size='large' shape='round' block>See all Activities</Button>
+        <Button type='secondary' href='/a/offer' size='large' shape='round' block>See all Activities</Button>
```
avowkind, line 87: This should be wrapped in <Link rather than have an href in the button. 

also text is not translated

Eteroa123: oh okay i'll wrap it in Link. 

@avowkind so i didn't make that change because that is another ticket . I want to follow the ticket and when I have done this i'll make the change. But if you want me to make all the changes on this ticket even though it not in this ticket then I can do that. But I would need to change that ticket and ad it  as a subtask.

Eteroa123: @avowkind  by wraping it in <Link> would that make it translated ??

PR #976, opened 2020-05-05
VP-1779 grab address using address selection tool

StellaRu:

## Proposed Changes? 🤔
1. Replaced current location selection box with a more specific address selection by addressFinder API

## Additional Info.🧐

Any additional info goes here

### Original
image

### Updated
image

image


components/Op/OpFormLocation.js
```diff
@@ -1,44 +1,184 @@
...
+      mounted.current = false
+      window.document.body.removeChild(script)
+    }
+  }, [])
+
+  useEffect(() => {
+    if (scriptLoaded && mounted.current) {
+      const widget = new window.AddressFinder.Widget(
+        addressRef.current.input,
+        'RGE83VMK4UYHXNJBP69W', // ADDRESSFINDER_KEY
```
avowkind, line 107: Is this a private key for our organisation?  If so it should not be put literally into the code base or sent to to the client browser. It would have to be in the env file for the production server and you will have to send the api request first to our server and then onto address finder.

components/Op/OpShortForm.js
```diff
@@ -43,15 +43,16 @@ class OpShortForm extends Component {
           // op.subtitle = values.subtitle
           // op.tags = values.tags
           op.duration = values.duration
-          op.locations = values.locations
-          delete op.location
+          // op.locations = values.locations
```
avowkind, line 46: we still should set the general location based on the address. otherwise the location based recommendations won't work.

StellaRu: Yes, but which kind of data I should set? the suburb level? or the city level? or the region level data?

components/Op/OpFormLocation.js
StellaRu: Yes, what's the env file we are using? I notice several files like beta.env.enc, alpha.env.encn or development.env.enc....

components/Op/OpShortForm.js
avowkind: City and region

StellaRu:
> great work thanks

I found that I made a mistake with the op.locations, now I fix it.

PR #979, opened 2020-05-07
VP-1687: count only active ops

richchurcher:

## Proposed Changes? 🤔
1. `getOpsForActivity` should not include DRAFT status ops in its count.


server/api/activity/__tests__/activity.lib.spec.js
```diff
@@ -52,7 +52,7 @@ const makeOp = (fromActivity, index) => {
     duration: fromActivity.duration,
     location: 'Northland',
     type: coin(OpportunityType.ASK, OpportunityType.OFFER),
-    status: coin(OpportunityStatus.DRAFT, OpportunityStatus.ACTIVE),
+    status: OpportunityStatus.ACTIVE,
```
avowkind, line 55: I suspect that this test change means that the fixture has no DRAFT ops so it will be difficult to prove that none are returned in the count.

richchurcher: Yep fair call.

PR #991, opened 2020-05-22
VP-1661 simplify DatePicker on OpAskForm and OpOfferForm

StellaRu:

## Proposed Changes? 🤔
1. Replaced DatePicker with RangePicker component
2. Removed time option

## Additional Info.🧐

Any additional info goes here

### Original
image

image

### Updated
image


components/Op/OpAskFormLong.js
```diff
@@ -195,9 +177,9 @@ class OpAskFormLong extends Component {
       this.props.form.validateFields((err, values) => {
         if (!err) {
           const op = this.props.op
-          const { startDateValue, endDateValue } = this.state
+          const { dateRangeValue } = this.state
           op.date = [] // Dirty work around to not change schema
```
avowkind, line 181: This line is not needed any more as you set date to an array in the next line.  It was only needed when we used push.

StellaRu: Fixed in https://github.com/voluntarily/vly2/pull/991/commits/b7dceb659582028ddbff88aad5885593db71201c

PR #1006, opened 2020-06-25
VP-1847: Rework activity and opportunity durations to use ISO 8601 Duration String

jordansimsmith:

## Proposed Changes? 🤔
1. Durations are currently stored as free text (4 hours, 4h, 4, etc.). To process these durations for the data dasboard metrics, they need to be machine readable. We have changed the durations to be stored as a [ISO 8601 Duration string](https://en.wikipedia.org/wiki/ISO_8601#Durations) (e.g. PT5H25M).
2. Duration form inputs on `OpShortForm` and `ActDetailForm` have been updated to input numbers for hours and minutes (which are then converted to ISO 8601 Duration String)
3. Components displaying the durations have been updated to handle the ISO String if needed, or render the free text to provide compatibility with existing database documents. 

## Additional Info.🧐
Legacy forms `OpAskFormLong` and `OpOfferFormLong` have not been updated as they have not been used.

### Original
ss

### Updated
 ss


waltzaround:
Tried testing on local - doesn't look like it's saving larger values correctly? Will try reproducing on other machines in a mo

HarrisonLeach1:
> Tried testing on local - doesn't look like it's saving larger values correctly? Will try reproducing on other machines in a mo
> 
> It returned this as well:
> 
> ```
>  `Warning: Failed prop type: Invalid prop `me._id` of type `object` supplied to `OpShortForm`, expected `string`.
>     in OpShortForm
>     in Form(OpShortForm) (at opdetailpage.js:175)
>     in div
>     in VTheme__FullPage (at opdetailpage.js:171)
>     in OpDetailPage
> ```

Thanks for spotting this @waltzaround, I'll make a fix soon, if you put in >= 24 hours it creates a day in the ISO string 
e.g. 25h and 10m is converted to "PT1D1H10M" but when we convert it back, we don't read the day value so it's converted to 1h 10m. Similar for >= 60 minutes.

waltzaround:
Imho it's fine to convert it to day, provided we render it on the listing - but happy with whichever approach you choose to take 🔥

HarrisonLeach1:
Hey @waltzaround in our case the duration prop is still a string of text (the ISO 8601 string), it's only split into hours and minutes upon rendering. 

The error looks like it's from `me._id`, I believe it's an existing issue, I'm not sure why it's not a string since it's an id.

 
I can change it to an object to remove the warning, this approach seems to also be used on the `persondetailpage.js`, otherwise we can leave it? what do you think?

PR #1011, opened 2020-07-01
VP-1842: Location and activity tag endpoints for the statistics API

HarrisonLeach1:
- [x] Run `npm run lint`, and there are no warnings.

## Proposed Changes? 🤔
1. Implement `GET api/statistics/locations` endpoint which retrieves the number of opportunity attendances by the organisation's volunteers at each location.
2. Implement `GET api/statistics/activityTags` endpoint which retrieves the weighted occurrences of each activity tag in the opportunities attended (see below for more info).
3. Unit tests added for each endpoint.

## Additional Info.🧐
The same query from #1010 is used, so it retrieves stats for a specified org, within the specified timeframe. 

The locations API schema is `GET /api/statistics/locations/:orgId/:timeframe` and returns an object similar to:
```json
[
  {"name": "auckland", "value": 2},
  {"name": "wellington", "value" : 1},
]
``` 
This means there have been 2 opportunity attendances in Auckland and 1 in Wellington. 
**Note**: the location used is the city of the opportunity. i.e. `locations[0]`

The activity tags API schema is `GET /api/statistics/activityTags/:orgId/:timeframe`, 

Given a set of attended opportunities with the following tags:
```json
[
  {"name": "Learning React", "tags": ["React", "JavaScript"]},
  {"name": "Learning Node.js", "tags": ["Node", "JavaScript"]},
  {"name": "Learning Python", "tags": ["Python"]}
]
```

This endpoint would return:

```json
[
  {"name": "react", "value" : 0.5},
  {"name": "javascript", "value": 1},
  {"name": "node", "value": 0.5},
  {"name": "python", "value": 1}
]
``` 
The value property is the sum of the weighted values of each tag.
The weighted value for a tag in an opportunity is: _1 / (number of tags in that opportunity)._  
This approach was decided to avoid opportunities with many tags skewing the statistics in a pie graph.

### DIFF
--- a/server/api/statistics/statistics.controller.js
+++ b/server/api/statistics/statistics.controller.js
@@ -54,6 +54,60 @@ const getSummary = async (req, res) => {
   }
 }
 
+const getLocations = async (req, res) => {
+  const { orgId, timeframe } = req.params
+
+  // authentication
+  const currentUser = req.session.me
+  if (!currentUser || !req.session.isAuthenticated) {
+    return res.status(401).send()
+  }
+
+  // authorisation
+  if (!Array.isArray(currentUser.role) || !currentUser.role.includes(Role.ORG_ADMIN) || !Array.isArray(currentUser.orgAdminFor) || !currentUser.orgAdminFor.includes(orgId)) {
+    return res.status(403).send()
+  }
+
+  let afterDate
+  try {
+    afterDate = parseStatisticsTimeframe(timeframe)
+  } catch (e) {
+    return res.status(400).send({ message: e.message })
+  }
+
+  try {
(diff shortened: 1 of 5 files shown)


server/api/statistics/statistics.controller.js
```diff
@@ -59,6 +54,123 @@ const getSummary = async (req, res) => {
   }
 }
 
+const getLocations = async (req, res) => {
+  const { orgId, timeframe } = req.params
+
+  // authentication
```
avowkind, line 60: You use this authentication and authorisation pattern fairly frequently in your api endpoints and may want to factor it out.

PR #1012, opened 2020-07-07
VP-1856: Implement data dashboard (/statistics) page

jordansimsmith:
## I do solemnly swear that I have:
- [x] Run `npm test` and all the tests pass.
- [x] Run `npm run lint` and there are no warnings.
- [x] Not decreased the overall test coverage?
- [x] Included the Jira ID and description in the PR title

## Proposed Changes? 🤔
1. Add /statistics page for org admins, where they can see summary, location and activity tag reports for the last month/year.
2. Add unit tests for react components.

## Additional Info.🧐
- Looking for feedback on the pie chart colours. Please comment any suggestions on the number of colours and which colours to use.
- No links/navigation to the statistics page are added yet, will add once we are happy to expose it formally.
- You can visit the page at `localhost:3122/statistics`, when logged in as an orgadmin
- Wireframes/Prototypes: https://www.figma.com/file/MU9xFHIxhJW9bErc6P1i9I/Voluntarily-Application?node-id=1265%3A5923

## Screenshots 📷
 
2020-07-07-154432_2560x1440_scrot

PR #1014, opened 2020-07-09
VP-1856: Refactor statistics page and API to be more maintainable

HarrisonLeach1:

## Proposed Changes? 🤔
1. Refactor `OrgStatisticsPage` to use `useSelector()` and `useDispatch()` instead of `connect`
2. Refactor statistics API auth into reusable middleware
3. Refactor reports to use a reusable container style 

## Additional Info.🧐
The changes are based on comments from #1012 and #1011.

For the `StatisticsLocationsReport` and `StatisticsActivityTagsReport` (the two pie graphs) we have decided not to refactor them into a reusable component because this would make them less flexible for extensions in the future. 

For example, we were thinking of potentially making the locations report into a two-level pie graph with the first level being region and second broken into suburbs. The reusable component would make this addition more difficult.


components/Statistics/OrgStatisticsTabs.js
```diff
@@ -1,41 +1,38 @@
...
-  }, [props.orgAdminFor])
+  }, [orgAdminFor])
 
-  if (!props.organisations.sync) {
+  if (!props.organisations || !props.organisations.sync) {
     return <ReduxLoading entity={props.organisations} label='organisations' />
   }
 
-  const orgs = props.organisations.data
+  const orgs = props.organisations.data || []
```
avowkind, line 25: under what condition will the second part of this line be executed?  .data must be falsy an empty array is not false and .data should not be undefined.

PR #1019, opened 2020-07-17
VP-1861: Implement feedback API

jordansimsmith:

## Proposed Changes? 🤔
1. Created feedback API
2. Created Feedback mongoose schema
3. Fixed minor fault in opportunity controller (nDeleted is not the correct property, changed to deletedCount)

## Additional Info.🧐
- Mongoose crudify was not used. From what it looks like, mongoose crudify offers no benefit to the existing controllers, as the default implementations are all overridden.
- Feedback documents are stored in a separate collection. See `v-students` for justifications.
- In the event that a unauthorised user requests a feedback resource by id, a 404 is returned (similar to other areas in the codebase). A 404 is returned instead of a 403, as to not provide information that this resource exists (but you do not have permission to it) to unauthorised and potentially malicious users.
- The feedback schema is currently minimal, supporting only the star rating entry. It is intended that this schema will be extended as further information is gathered.
- CASL framework used for role based authorisation.


jordansimsmith:
> looks good. I'm wondering whether it is possible for a person to add feedback to an opportunity that they didn't have anything to do with. Probably there would need to be a valid interest or interestArchive record linking the person to the opportunity. As a black hat would I be able to leave negative feedback on every opportunity in the list for example? Not a major security hole but it would be good to be clear on the business rule about who can leave feedback and when i.e. only when the op is completed. Can you leave feedback on a cancelled op?

